PR #16, opened 2020-08-26
solution

aliceblunt8:
- [DEMO LINK](https://aliceblunt8.github.io/js_sort_table_DOM/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -10,10 +10,10 @@
     <table>
       <thead>
         <tr>
-          <th>Name</th>
-          <th>Position</th>
-          <th>Age</th>
-          <th>Salary</th>
+          <th data-type="string">Name</th>
+          <th data-type="string">Position</th>
+          <th data-type="number">Age</th>
+          <th data-type="currency">Salary</th>
         </tr>
       </thead>
       <tbody>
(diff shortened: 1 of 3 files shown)


src/index.html
```diff
@@ -10,10 +10,10 @@
...
       <thead>
         <tr>
-          <th>Name</th>
-          <th>Position</th>
-          <th>Age</th>
-          <th>Salary</th>
+          <th data-type="string">Name</th>
+          <th data-type="string">Position</th>
+          <th data-type="number">Age</th>
+          <th data-type="currency">Salary</th>
```
krajnova, line 16: according to the task you mustn't change the html file

src/scripts/main.js
```diff
@@ -1,3 +1,42 @@
...
+        toNumber(a.cells[columnIndex].textContent)
+        - toNumber(b.cells[columnIndex].textContent);
+  }
+
+  rows.sort(compareFunction);
+  tbody.append(...rows);
+}
+
+function toNumber(str) {
+  return +str.replace('$', '').replace(',', '');
```
krajnova, line 41: you could use regExp here instead of using two replaces, but it isn't necessary)

krajnova (changes requested):

krajnova (approved):
your code looks better now)

irenhh (approved):

PR #21, opened 2020-08-26
Solution

uni-8:
- [DEMO LINK](https://uni-8.github.io/js_sort_table_DOM/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,23 @@
 'use strict';
 
-// write code here
+const table = document.querySelector('table');
+
+table.tHead.addEventListener('click', (event) => {
+  sortList(event.target.cellIndex);
+});
+
+function sortList(index) {
+  const sortedRows = [...table.tBodies[0].rows].sort((a, b) => {
+    let first = a.cells[index].textContent;
+    let second = b.cells[index].textContent;
+
+    if (index === 3) {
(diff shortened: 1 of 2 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,23 @@
...
+table.tHead.addEventListener('click', (event) => {
+  sortList(event.target.cellIndex);
+});
+
+function sortList(index) {
+  const sortedRows = [...table.tBodies[0].rows].sort((a, b) => {
+    let first = a.cells[index].textContent;
+    let second = b.cells[index].textContent;
+
+    if (index === 3) {
```
aliceblunt8, line 14: it's better to find a universal condition to identify a necessary column (new columns can be added and this index will change)

aliceblunt8 (approved):
look good

src/scripts/main.js
uni-8: yes, your right. but the title text can change too. i can't think of a universal condition here for all cases

Denchepornyuk (approved):

src/scripts/main.js
aliceblunt8: yes, agree. but this case was too obvious to ignore ;)

Denchepornyuk (approved):

PR #36, opened 2020-08-27
first solution

leonbohdan:
- [DEMO LINK](https://leonbohdan.github.io/js_sort_table_DOM/)


furude-rika (approved):
you've done a good alternative solution!
small comment: your functions 'sortByName' and 'sortByPosition' are the same, maybe it's better to remove one

src/scripts/main.js
```diff
@@ -1,3 +1,60 @@
...
+    th.dataset.name = th.innerText;
+  });
+
+function sortByName(a, b) {
+  return a > b ? 1 : -1;
+}
+
+function sortByPosition(a, b) {
+  return a > b ? 1 : -1;
+}
```
irenhh, line 17: what is the difference?

src/scripts/main.js
```diff
@@ -1,3 +1,60 @@
 'use strict';
 
-// write code here
+const tableHead = document.querySelector('thead');
+const tableBody = document.querySelector('tbody');
+
+[...document.querySelectorAll('th')]
+  .map(th => {
+    th.dataset.name = th.innerText;
```
irenhh, line 8: do not do any side effects in map

src/scripts/main.js
leonbohdan: I think about this and I'll fix it

src/scripts/main.js
leonbohdan: changed by forEach loop

irenhh (approved):

PR #66, opened 2020-10-17
solution

zkhomer:
[DEMO LINK](https://zkhomer.github.io/js_sort_table_DOM/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,25 @@
 'use strict';
 
-// write code here
+const list = document.querySelector('tbody');
+
+const listHeader = [ ...document.querySelectorAll('th') ];
+// console.log(listHeader)
+listHeader.map((item, index) => {
+  // debugger
+  item.addEventListener('click', () => {
+    const rows = [ ...list.querySelectorAll('tr') ];
+
+    rows.sort((a, b) => {
+      const itemA = a.children[index].innerText.replace(/[^A-Z0-9]/gi, '');
+      const itemB = b.children[index].innerText.replace(/[^A-Z0-9]/gi, '');
+
+      if (!isNaN(itemA)) {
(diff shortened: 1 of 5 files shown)


Andrii1996 (approved):

src/scripts/main.js
```diff
@@ -1,3 +1,25 @@
 'use strict';
 
-// write code here
+const list = document.querySelector('tbody');
```
aliceblunt8, line 3:
```suggestion
const tbody = document.querySelector('tbody');
```

src/scripts/main.js
```diff
@@ -1,3 +1,25 @@
...
+
+const listHeader = [ ...document.querySelectorAll('th') ];
```
aliceblunt8, line 5: you can find table header like this -> tbody.tHead
read more about different props connected to tables
https://learn.javascript.ru/dom-navigation#dom-navigation-tables

src/scripts/main.js
```diff
@@ -1,3 +1,25 @@
...
+// console.log(listHeader)
+listHeader.map((item, index) => {
+  // debugger
+  item.addEventListener('click', () => {
```
aliceblunt8, line 9: let's put an event listener to the whole table header (use event delegation for this)

src/scripts/main.js
```diff
@@ -1,3 +1,25 @@
...
+    const rows = [ ...list.querySelectorAll('tr') ];
```
aliceblunt8, line 10:
```suggestion
    const rows = [ ...tbody.rows];
```

src/scripts/main.js
aliceblunt8: if you still want to know the index of each heading, use `td.cellIndex` prop

aliceblunt8 (changes requested):

PR #67, opened 2020-10-17
solution

andlap24:
[DEMO LINK](https://andlap24.github.io/js_sort_table_DOM/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,50 @@
 'use strict';
 
-// write code here
+const table = document.querySelector('table');
+const tableHead = [...table.rows[0].cells];
+const tableRows = [...table.tBodies[0].rows];
+
+table.addEventListener('click', (event) => {
+  if (tableHead.includes(event.target)) {
+    const sortString = (a, b) => {
+      return a.localeCompare(b);
+    };
+
+    const sortAge = (a, b) => {
+      return +a - +b;
+    };
+
+    const sortSalary = (a, b) => {
+      return +(a.replace(/[\s.,$]/g, '')) - +(b.replace(/[\s.,$]/g, ''));
+    };
+
+    const selectedRow = event.target.textContent;
+
+    if (selectedRow === 'Name' || selectedRow === 'Position') {
+      return sortTable(event.target, sortString);
+    }
+
+    if (selectedRow === 'Age') {
+      return sortTable(event.target, sortAge);
+    }
+
+    if (selectedRow === 'Salary') {
+      return sortTable(event.target, sortSalary);
+    }
+  }
+});
+
+function sortTable(columnHeader, compareFunction) {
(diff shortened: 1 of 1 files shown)


goncharukop (approved):

src/scripts/main.js
```diff
@@ -1,3 +1,50 @@
 'use strict';
 
-// write code here
+const table = document.querySelector('table');
+const tableHead = [...table.rows[0].cells];
+const tableRows = [...table.tBodies[0].rows];
+
+table.addEventListener('click', (event) => {
```
aliceblunt8, line 7: let's assign event listener just on the table header

src/scripts/main.js
```diff
@@ -1,3 +1,50 @@
 'use strict';
 
-// write code here
+const table = document.querySelector('table');
+const tableHead = [...table.rows[0].cells];
```
aliceblunt8, line 4: if you want to find table head just use table.tHead

src/scripts/main.js
```diff
@@ -1,3 +1,50 @@
...
+
+  const sortedRows = tableRows.sort((a, b) => {
+    const columnA = a.querySelectorAll('td')[index].textContent;
+    const columnB = b.querySelectorAll('td')[index].textContent;
+
+    return compareFunction(columnA, columnB);
+  });
+
+  for (const row of sortedRows) {
+    table.tBodies[0].append(row);
```
aliceblunt8, line 48: instead of for-loop you can use spread operator


```suggestion
    table.tBodies[0].append(...row);
```

src/scripts/main.js
```diff
@@ -1,3 +1,50 @@
...
+function sortTable(columnHeader, compareFunction) {
+  const index = tableHead.indexOf(columnHeader);
+
+  const sortedRows = tableRows.sort((a, b) => {
+    const columnA = a.querySelectorAll('td')[index].textContent;
+    const columnB = b.querySelectorAll('td')[index].textContent;
+
+    return compareFunction(columnA, columnB);
+  });
+
```
aliceblunt8, line 46: cool! I like that you create a separate function for sorting, but let's refactor this function a little bit
instead of creating and then passing compare function for each single case, let move all the checks inside the  sortTable func

aliceblunt8 (changes requested):

src/scripts/main.js
andlap24: unfortunately, i don't understand exactly( . Move all if statements into sortTable function? And throw into sortTable arguments name of clicked column name?

andlap24: ok, i think i find the way how to solve this task)

aliceblunt8 (approved):
the idea with sortTable function was great! I meant there was no need to create additional sortString, sortAge, sortSalary. anyway, this solution is also ok, approved)

PR #71, opened 2020-10-19
Solution

Feuerbach98:
- [DEMO LINK](https://feuerbach98.github.io/js_sort_table_DOM/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,34 @@
 'use strict';
 
-// write code here
+const table = document.querySelector('table');
+const tableHead = [...table.tHead.firstElementChild.cells];
+const tableRows = [...table.tBodies[0].rows];
+
+table.addEventListener('click', (event) => {
+  if (tableHead.includes(event.target)) {
+    const sortString = (a, b) => {
+      return b < a ? 1 : -1;
+    };
+
+    const sortNumber = (a, b) => {
+      return +(a.replace(/[^+\d]/g, '')) - +(b.replace(/[^+\d]/g, ''));
+    };
+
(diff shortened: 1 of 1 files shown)


goncharukop (approved):

src/scripts/main.js
```diff
@@ -1,3 +1,34 @@
...
+
+    return (selectedRow === 'Age' || selectedRow === 'Salary')
+      ? sortTable(event.target, sortNumber)
+      : sortTable(event.target, sortString);
+  }
+});
+
+function sortTable(header, compareFunction) {
+  const indexOfColumn = tableHead.indexOf(header);
+
```
aliceblunt8, line 27: I like that you create a separate function for sorting, but let's refactor this function a little bit - instead of creating and then passing compare function for each single case (sortString, sortNumber), let move all the checks inside the sortTable function

src/scripts/main.js
```diff
@@ -1,3 +1,34 @@
 'use strict';
 
-// write code here
+const table = document.querySelector('table');
+const tableHead = [...table.tHead.firstElementChild.cells];
+const tableRows = [...table.tBodies[0].rows];
+
+table.addEventListener('click', (event) => {
```
aliceblunt8, line 7: put event listener on table head

aliceblunt8: then you don't need to check whether the table head contains the event.target

src/scripts/main.js
aliceblunt8: this way, we'll have a single function responsible for table sorting

aliceblunt8 (changes requested):

PR #119, opened 2021-01-06
solution

xipholena:
[DEMO LINK](https://xipholena.github.io/js_sort_table_DOM/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,46 @@
 'use strict';
 
 // write code here
+
+const people = document.querySelector('tbody').children;
+
+const titles = document.querySelectorAll('thead > tr > th');
+const [fullName, position, age, salary] = titles;
+
+fullName.classList.add('full-name');
+position.classList.add('position');
+age.classList.add('age');
+salary.classList.add('salary');
+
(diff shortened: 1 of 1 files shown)


sasha39612 (approved):

src/scripts/main.js
```diff
@@ -1,3 +1,46 @@
...
+    document.querySelector('tbody').append(...sortedByAge);
+  }
+
+  if (e.target.classList.contains('salary')) {
+    const sortedBySalary = [...people].sort((a, b) => +(a.cells[3].innerText)
+      .replace(/\$/, '').replace(/,/, '') - +(b.cells[3].innerText)
+      .replace(/\$/, '').replace(/,/, ''));
+
+    document.querySelector('tbody').append(...sortedBySalary);
+  }
```
varseniuk, line 45: You are repeating yourself too much. Also, there's no need to use the `sort` method four times. You can only use method `sort` one time and inside it check, what you are trying to sort (string, number, or number with $-sign)

varseniuk (changes requested):
Logic is correct, but DRY and use `sort` method only once

varseniuk (approved):

PR #122, opened 2021-01-06
Added sorting event

vorfolomeevav:
- [DEMO LINK](https://vorfolomeevav.github.io/js_sort_table_DOM/)


dtsehelnyk (approved):

varseniuk (approved):
The logic is correct. Code is laconic and you don't repeat yourself, but It will be great if you will use event delegation instead of creating listeners for every table cell

PR #126, opened 2021-01-07
Sort table

tamara-ostapets:
- [DEMO LINK](https://tamara-ostapets.github.io/js_sort_table_DOM/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,48 @@
 'use strict';
 
-// write code here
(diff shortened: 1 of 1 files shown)


Lara-l-art (approved):
Good job!

varseniuk (approved):
Logic is correct, but instead of using the `sort` method three times you could use it only once and inside it check it's arguments (whether they are numbers, strings or numbers with $-sign)

PR #150, opened 2021-01-12
Solution

MaximZhuravlov:
- [DEMO LINK](https://MaximZhuravlov.github.io/js_sort_table_DOM/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,29 @@
 'use strict';
 
-// write code here
+const tableHead = document.querySelector('table thead');
+const tableBody = document.querySelector('table tbody');
+const tableRows = [...document.querySelectorAll('table tbody tr')];
+
+tableHead.addEventListener('click', e => {
+  const criterion = [...tableHead.children[0].children].indexOf(e.target);
+
+  tableRows.sort((a, b) => {
+    const textContentA = a.children[criterion].textContent;
+    const textContentB = b.children[criterion].textContent;
+
+    if (criterion === 2) { // sorting by age
+      return Number(textContentA) - Number(textContentB);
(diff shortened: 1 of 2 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,29 @@
 'use strict';
 
-// write code here
+const tableHead = document.querySelector('table thead');
+const tableBody = document.querySelector('table tbody');
+const tableRows = [...document.querySelectorAll('table tbody tr')];
```
aliceblunt8, line 5: you can use table selectors, like tBodies, tHead, rows, etc
https://learn.javascript.ru/dom-navigation#dom-navigation-tables

src/scripts/main.js
```diff
@@ -1,3 +1,29 @@
...
+
+tableHead.addEventListener('click', e => {
+  const criterion = [...tableHead.children[0].children].indexOf(e.target);
```
aliceblunt8, line 8: use .cellIndex
https://learn.javascript.ru/dom-navigation#dom-navigation-tables

src/scripts/main.js
```diff
@@ -1,3 +1,29 @@
...
+
+  tableRows.sort((a, b) => {
+    const textContentA = a.children[criterion].textContent;
+    const textContentB = b.children[criterion].textContent;
+
+    if (criterion === 2) { // sorting by age
```
aliceblunt8, line 14: switch case

aliceblunt8: also, it's not the best idea to sort table depending on cellIndex, cause they may change, so it's better to come up with more universal solution

aliceblunt8 (approved):

PR #212, opened 2021-04-04
solution

Dinexpod:
[DEMO LINK](https://Dinexpod.github.io/js_sort_table_DOM/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,41 @@
 'use strict';
 
-// write code here
+const theadRows = document.querySelector('thead tr');
+const tbodyElement = document.querySelector('tbody');
+
+theadRows.addEventListener('click', e => {
+  const trElements = [];
+  let columnNumber;
+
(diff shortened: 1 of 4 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,41 @@
...
+
+  for (const trChild of tbodyElement.children) {
+    trElements.push(trChild);
+  }
+
+  trElements.sort((a, b) => {
+    let aData = a.children[columnNumber].textContent;
+    let bData = b.children[columnNumber].textContent;
+
+    if (columnNumber === 3) {
```
vkryvytskyy, line 24: better to use event.target.name, to understand which column was clicked and which data you should pick to apply sorting. Indices are not stable for this purpose. If I move `salary` column left, there will be `age` under columnNumber === 3, and your sorting wouldn't work properly

vkryvytskyi (approved):

PR #279, opened 2021-12-24
done

VOV44ER:
https://vov44er.github.io/js_sort_table_DOM/

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,31 @@
 'use strict';
 
-// write code here
+const tbody = document.querySelector('tbody');
+
+const sort = (index) => {
+  let sorted = null;
+
+  if (index === 3) {
+    sorted = [...tbody.rows].sort(
+      (a, b) =>
+        a.cells[index].innerHTML.replace(/[^0-9]/g, '')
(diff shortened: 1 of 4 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,31 @@
...
+document.addEventListener('click', (eve) => {
+  if (eve.target.innerText === 'Name') {
+    sort(0);
+  } else if (eve.target.innerText === 'Position') {
+    sort(1);
+  } else if (eve.target.innerText === 'Age') {
+    sort(2);
+  } else if (eve.target.innerText === 'Salary') {
+    sort(3);
+  }
```
aliceblunt8, line 30: this looks weird, move these checks inside sort function and check everything there

aliceblunt8 (changes requested):

src/scripts/main.js
```diff
@@ -1,3 +1,43 @@
...
+      break;
+
+    case 'Age':
+      index = 2;
+      break;
+
+    case 'Salary':
+      index = 3;
+      break;
+  }
```
aliceblunt8, line 24: remove index at all, write all logic inside cases (you can combine them as well if you need to do this)

aliceblunt8 (changes requested):

PR #301, opened 2022-03-09
Solution

dima-cherednyk:
- [DEMO LINK](https://dima-cherednyk.github.io/js_sort_table_DOM/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,51 @@
 'use strict';
 
-// write code here
+const body = document.querySelector('tbody');
+const person = body.querySelectorAll('tr');
+const head = document.querySelector('thead');
+
+head.addEventListener('click', (e) => {
+  const order = e.target;
+
+  if (order.textContent === 'Name') {
+    const y = [...person].map(x => x);
+    const res = y.sort((a, b) =>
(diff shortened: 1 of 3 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,51 @@
...
-// write code here
+const body = document.querySelector('tbody');
+const person = body.querySelectorAll('tr');
+const head = document.querySelector('thead');
+
+head.addEventListener('click', (e) => {
+  const order = e.target;
+
+  if (order.textContent === 'Name') {
+    const y = [...person].map(x => x);
```
vitaliy-martynyuk, line 11:
```suggestion
    const y = [...person];
```

src/scripts/main.js
```diff
@@ -1,3 +1,51 @@
...
+    const res = y.sort((a, b) =>
```
vitaliy-martynyuk, line 12: `sort` is a mutating method, so there is no need to create new variable for sorted array
```suggestion
    y.sort((a, b) =>
```

vitaliy-martynyuk (changes requested):

vitaliy-martynyuk (approved):

PR #304, opened 2022-03-17
Sort table solution

MykhailoIvchenko:
https://mykhailoivchenko.github.io/js_sort_table_DOM/

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,72 @@
 'use strict';
 
-// write code here
+let clickedColumn;
+
+const tableArray = [];
+
+function toNumber(numberAsString) {
+  let resultString = numberAsString;
+
+  if (numberAsString[0] === '$') {
+    resultString = numberAsString.slice(1);
+
+    if (typeof (parseInt(resultString.split(',').join('')))
+    === 'number') {
+      resultString = parseInt(resultString.split(',').join(''));
+    }
+  } else if (parseInt(resultString)) {
+    resultString = parseInt(numberAsString);
+  }
+
+  return resultString;
+}
+
+function sortByColumn(tableToSort, columnNumber) {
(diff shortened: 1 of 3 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,72 @@
...
+    if (typeof (parseInt(resultString.split(',').join('')))
+    === 'number') {
+      resultString = parseInt(resultString.split(',').join(''));
+    }
+  } else if (parseInt(resultString)) {
+    resultString = parseInt(numberAsString);
+  }
+
+  return resultString;
+}
```
Bohdan-Maksymiuk, line 22: use Number() or Unary Operator (+) instead of this funciton
https://dev.to/sanchithasr/7-ways-to-convert-a-string-to-number-in-javascript-4l

src/scripts/main.js
```diff
@@ -1,3 +1,72 @@
...
+
+function sortByColumn(tableToSort, columnNumber) {
+  const tableRows = tableToSort.querySelector('tbody').rows;
+  const tableHeader = tableToSort.querySelector('thead');
+  const tableFooter = tableToSort.querySelector('tfoot');
```
Bohdan-Maksymiuk, line 27:
```suggestion
const list = document.querySelector('tbody');
const row = list.querySelectorAll('tr');
const head = document.querySelector('thead');
```

Bohdan-Maksymiuk (changes requested):

src/scripts/main.js
MykhailoIvchenko: Number() or Unary Opertor doesn't work with such string as $162,700... I can use them only for Age column...
I can shorten toNumber function to use it only with string similar to $162,700. But in this case it would be a lot of checks in other parts of the code.
Maybe, I've understood something incorrect.

PR #309, opened 2022-04-06
Sort Table DOM

Vladyslav317:
https://vladyslav317.github.io/js_sort_table_DOM/

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,32 @@
 'use strict';
 
-// write code here
+const tbody = document.querySelector('tbody');
+const rows = document.querySelector('tbody').rows;
+const table = document.querySelector('table');
+
+function getToNumber(str) {
+  return str.split('$').join('').split(',').join('');
+}
+
+table.addEventListener('click', e => {
+  const targeted = e.target.closest('th');
+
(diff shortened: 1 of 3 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,32 @@
 'use strict';
 
-// write code here
+const tbody = document.querySelector('tbody');
+const rows = document.querySelector('tbody').rows;
```
aliceblunt8, line 4:
```suggestion
const tbody = document.querySelector('tbody');
const rows = tbody.rows;
```

src/scripts/main.js
```diff
@@ -1,3 +1,32 @@
...
+function getToNumber(str) {
+  return str.split('$').join('').split(',').join('');
+}
+
+table.addEventListener('click', e => {
+  const targeted = e.target.closest('th');
+
+  if (!targeted) {
+    return;
+  }
```
aliceblunt8, line 16: or you can add a listener directly to thead, then you won't need this check

aliceblunt8 (changes requested):

Bohdan-Maksymiuk (approved):

PR #350, opened 2022-07-06
sorted table

silchyn:
[DEMO](https://silchyn.github.io/js_sort_table_DOM/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,42 @@
 'use strict';
 
-// write code here
+const headRow = document.querySelector('thead').firstElementChild;
+const tbody = document.querySelector('tbody');
+const data = [];
+
+for (const row of tbody.children) {
+  data.push([row.children[0].textContent,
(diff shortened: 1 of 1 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,42 @@
...
 
-// write code here
+const headRow = document.querySelector('thead').firstElementChild;
+const tbody = document.querySelector('tbody');
+const data = [];
+
+for (const row of tbody.children) {
+  data.push([row.children[0].textContent,
+    row.children[1].textContent, +row.children[2].textContent,
+    +row.children[3].textContent.replace(/\$|,/g, '')]);
```
VETALEK, line 10:
```suggestion
  data.push({
  	salary: ... ,
  	name: ... ,
  });
```

better use objects, so it will be more clear what we are sorting

VETALEK (changes requested):
GJ 👍 
mind using array of objects, not an array of arrays

VETALEK (approved):
👍

PR #357, opened 2022-07-12
Develop

WhiteAshDc:


src/scripts/main.js
```diff
@@ -1,3 +1,59 @@
...
+      case a < b:
+        return -1;
+      case a === b:
+        return 0;
+    }
+  });
+
+  [].forEach.call(rows, (row) => {
+    tbody.removeChild(row);
+  });
```
ArtemTopolskyi, line 48:
```suggestion
  [...rows].forEach((row) => {
    tbody.removeChild(row);
  });
```

Bohdan-Maksymiuk, line 48: couldn't we just iterate through rows?
https://stackoverflow.com/questions/16053357/what-does-foreach-call-do-in-javascript
read `Update for ES6(ES2015) and Beyond`

Bohdan-Maksymiuk (approved):

src/scripts/main.js
Bohdan-Maksymiuk: https://attacomsian.com/blog/javascript-convert-nodelist-to-array#:~:text=A%20NodeList%20is%20an%20array,()%20on%20a%20NodeList%20object.

Bohdan-Maksymiuk: better use spread

PR #400, opened 2022-09-04
solution

Andrii-Kuzmenko:
[DEMO LINK](https://Andrii-Kuzmenko.github.io/js_sort_table_DOM/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,43 @@
 'use strict';
 
-// write code here
+const tableHeaders = document.querySelector('thead');
+const tableBody = document.querySelector('tbody');
+
+const tableData = [...tableBody.children].map(tableRow => ({
+  name: tableRow.children[0].innerText,
+  position: tableRow.children[1].innerText,
+  age: tableRow.children[2].innerText,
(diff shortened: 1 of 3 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,43 @@
...
+    if (event.target.innerText === 'Name') {
+      return element1.name.localeCompare(element2.name);
+    } else if (event.target.innerText === 'Position') {
+      return element1.position.localeCompare(element2.position);
+    } else if (event.target.innerText === 'Age') {
+      return element1.age - element2.age;
+    } else {
+      return +changeSalaryFormat(element1.salary)
+        - +changeSalaryFormat(element2.salary);
+    }
```
Bohdan-Maksymiuk, line 24: avoid such nesting `else if` conditions
This code is hard to read and maintain

Bohdan-Maksymiuk: better to use `switch case` in such situations

Bohdan-Maksymiuk (changes requested):

AlisaVlasova (approved):

PR #409, opened 2022-09-14
Solution by DK

DimaK123:
[DEMO LINK](https://DimaK123.github.io/js_sort_table_DOM/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,40 @@
 'use strict';
 
-// write code here
+const bodyOfTable = document.querySelector('tbody');
+
+const headOfTable = document.querySelector('tHead > tr');
+
+const tableRows = bodyOfTable.rows;
+
+headOfTable.addEventListener('click', () => {
+  let arrayByName;
+
+  switch (true) {
+    case event.target.cellIndex === 0:
+      arrayByName = [...tableRows].sort((a, b) => {
+        return a.children[0].innerText.localeCompare(b.children[0].innerText);
+      });
+      break;
+
+    case event.target.cellIndex === 1:
+      arrayByName = [...tableRows].sort((a, b) => {
+        return a.children[1].innerText.localeCompare(b.children[1].innerText);
+      });
+      break;
+
+    case event.target.cellIndex === 2:
+      arrayByName = [...tableRows].sort((a, b) =>
+        a.children[2].innerText - b.children[2].innerText);
+      break;
+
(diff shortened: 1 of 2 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,40 @@
...
+      arrayByName = [...tableRows].sort((a, b) => {
+        return a.children[0].innerText.localeCompare(b.children[0].innerText);
+      });
+      break;
+
+    case event.target.cellIndex === 1:
+      arrayByName = [...tableRows].sort((a, b) => {
+        return a.children[1].innerText.localeCompare(b.children[1].innerText);
+      });
+      break;
```
ivanchuk-yurii, line 23: you are using `switch case` as `if`, it's not correct. you can combine similar logic in one `case` and use `event.target.cellIndex` as index
```
const index = event.target.cellIndex;
switch (index) {
  case 0:
  case 1:
    a.children[index].localeCompare(b.children[index];
     ...
}
```

src/scripts/main.js
```diff
@@ -1,3 +1,40 @@
...
+
+    case event.target.cellIndex === 2:
+      arrayByName = [...tableRows].sort((a, b) =>
+        a.children[2].innerText - b.children[2].innerText);
+      break;
+
+    case event.target.cellIndex === 3:
+      arrayByName = [...tableRows].sort((a, b) => {
+        const first = parseFloat((a.children[3].innerText).slice(1));
```
ivanchuk-yurii, line 32: `parseFloat('138,575') = 138`, it's not that you want, choose another way of conversion

ivanchuk-yurii (changes requested):
think about creating helper function for conversion you do multiple times

DimaK123:
> think about creating helper function for conversion you do multiple times

I'm not sure that i've understanded correct, but i made a function and used it in event.

src/scripts/main.js
DimaK123: > you are using `switch case` as `if`, it's not correct. you can combine similar logic in one `case` and use `event.target.cellIndex` as index
> 
> ```
> const index = event.target.cellIndex;
> switch (index) {
>   case 0:
>   case 1:
>     a.children[index].localeCompare(b.children[index];
>      ...
> }
> ```

Done

src/scripts/main.js
DimaK123: > `parseFloat('138,575') = 138`, it's not that you want, choose another way of conversion

Done

PR #427, opened 2022-10-08
solution

serhii-yunnikov:
https://serhii-yunnikov.github.io/js_sort_table_DOM/

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,41 @@
 'use strict';
 
-// write code here
+const heading = document.querySelector('thead').firstElementChild;
+const headingItems = [...heading.querySelectorAll('th')];
+const tableBody = document.querySelector('tbody');
+const tableRows = [...tableBody.children];
+
+headingItems.forEach(item => item.setAttribute('class', 'column'));
+
+function tableSort(column) {
+  const index = headingItems.indexOf(column);
+
+  const callbackSort = (rowA, rowB) => {
+    const first = rowA.children[index].innerText;
+    const second = rowB.children[index].innerText;
+
+    if (index <= 1) {
+      return first.localeCompare(second);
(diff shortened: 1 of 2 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,41 @@
...
+headingItems.forEach(item => item.setAttribute('class', 'column'));
+
+function tableSort(column) {
+  const index = headingItems.indexOf(column);
+
+  const callbackSort = (rowA, rowB) => {
+    const first = rowA.children[index].innerText;
+    const second = rowB.children[index].innerText;
+
+    if (index <= 1) {
```
qoa11a, line 17: This is hard-coded solution. If I replace position of columns your code won't work as expected. Try to solve it another way.

src/scripts/main.js
```diff
@@ -1,3 +1,41 @@
...
+      return first.localeCompare(second);
+    } else if (index === 3) {
+      return first.slice(1).split(',').join('')
```
qoa11a, line 20: Create function for transforming this type of string into number as you repeat it two times at least.

qoa11a (changes requested):

src/scripts/main.js
```diff
@@ -1,3 +1,49 @@
...
+
+    const transformSalary = (cell) => {
+      return cell.slice(1).split(',').join('');
+    };
+
+    if (column.dataset.type === 'text') {
+      return first.localeCompare(second);
+    } else if (column.dataset.type === 'salary') {
+      return transformSalary(first)
+      - transformSalary(second);
```
Anastasiia-Svintsova, line 29: It's hard to read
```suggestion
      return transformSalary(first) - transformSalary(second);
```

Anastasiia-Svintsova (approved):
Good👍

PR #431, opened 2022-10-18
JS sort table DOM (first solution)

Oleksandr-Naumchak:
[DEMO LINK](https://Oleksandr-Naumchak.github.io/js_sort_table_DOM/)


src/scripts/main.js
```diff
@@ -1,3 +1,35 @@
...
+
+          return +stringBegin - +stringEnd;
+        }
+
+        return stringBegin.localeCompare(stringEnd);
+      });
+    }
+  }
+
+  for (const row of rows) {
```
TarasHoliuk, line 30: looks like the whole table are clickable after the first load because of this
it has to be fixed

src/scripts/main.js
```diff
@@ -1,3 +1,35 @@
 'use strict';
 
-// write code here
+const table = document.querySelector('table');
+let rows = document.querySelectorAll('tbody > tr');
+const headers = document.querySelectorAll('thead > tr > th');
+
+const sortTable = (elem) => {
+  for (let i = 0; i < 4; i++) {
+    if (elem.target === headers[i]) {
```
TarasHoliuk, line 9: first of all better to check if it's the correct tag and just then do sorting

TarasHoliuk (changes requested):

src/scripts/main.js
```diff
@@ -1,3 +1,44 @@
...
+};
+
+const sortTable = (elem) => {
+  const item = elem.target;
+
+  if (!item.matches('th')) {
+    return;
+  };
+
+  if (item.innerText === 'Name') {
```
AlisaVlasova, line 18: use switch case here

AlisaVlasova (changes requested):

nazarmatsevych (approved):

PR #432, opened 2022-10-18
solution

Moblat890:


src/scripts/main.js
```diff
@@ -1,3 +1,30 @@
...
+  if (e.target.tagName !== 'TH') {
+    return;
+  };
+
+  rows.sort((a, b) => {
+    const elementA = a.cells[index].textContent;
+    const elementB = b.cells[index].textContent;
+
+    if (e.target.textContent === 'Salary') {
+      return salaryToNumber(elementA) - salaryToNumber(elementB);
```
Anastasiia-Svintsova, line 19: It's not good to use a salaryToNumber function before it's declared.

Anastasiia-Svintsova (approved):

src/scripts/main.js
```diff
@@ -1,3 +1,30 @@
...
+const tbody = document.querySelector('tbody');
+const theadList = [...document.querySelector('thead tr').children];
+const rows = [...tbody.rows];
+
+document.body.addEventListener('click', e => {
+  const index = theadList.indexOf(e.target);
+
+  function salaryToNumber(sal) {
+    return +sal.slice(1).split(',').join('');
+  };
```
polosanya, line 12: It is better to move this function outside of eventListener, so that you can reuse it somewhere else if needed

src/scripts/main.js
```diff
@@ -1,3 +1,30 @@
 'use strict';
 
-// write code here
+const tbody = document.querySelector('tbody');
+const theadList = [...document.querySelector('thead tr').children];
+const rows = [...tbody.rows];
+
+document.body.addEventListener('click', e => {
```
polosanya, line 7: You can add eventListener not for the whole body, but only for table, or even its header

polosanya (changes requested):
Good job! Please check my comments

polosanya (approved):
Nice!

PR #436, opened 2022-10-22
Solution

JohnRust12:
- [DEMO LINK](https://JohnRust12.github.io/js_sort_table_DOM/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,48 @@
 'use strict';
 
-// write code here
+const thead = document.querySelector('thead');
+const tbody = document.querySelector('tbody');
+const persons = tbody.querySelectorAll('tr');
+
+thead.addEventListener('click', (e) => {
+  let sortedPersons;
+
+  switch (e.target.textContent) {
+    case 'Name':
+      sortedPersons = [...persons]
+        .sort((a, b) => {
+          return a.children[0].textContent
+            .localeCompare(b.children[0].textContent);
+        });
+
+      sortedPersons.forEach(person => tbody.append(person));
+      break;
+    case 'Position':
+      sortedPersons = [...persons]
+        .sort((a, b) => {
+          return a.children[1].textContent
+            .localeCompare(b.children[1].textContent);
+        });
+
+      sortedPersons.forEach(person => tbody.append(person));
+      break;
+    case 'Age':
+      sortedPersons = [...persons]
+        .sort((a, b) => {
+          return a.children[2].textContent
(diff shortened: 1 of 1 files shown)


stanislav-topikha (changes requested):
put `switch/case` inside of `sort` method, in this way it will be more compact

JohnRust12:
+

src/scripts/main.js
```diff
@@ -1,3 +1,48 @@
 'use strict';
 
-// write code here
+const thead = document.querySelector('thead');
+const tbody = document.querySelector('tbody');
+const persons = tbody.querySelectorAll('tr');
```
polosanya, line 5:
```suggestion
const people = tbody.querySelectorAll('tr');
```

src/scripts/main.js
```diff
@@ -1,3 +1,48 @@
...
+
+thead.addEventListener('click', (e) => {
+  let sortedPersons;
```
polosanya, line 8:
```suggestion
  let sortedPeople;
```

src/scripts/main.js
```diff
@@ -1,3 +1,48 @@
...
+      sortedPersons.forEach(person => tbody.append(person));
+      break;
+    case 'Salary':
+      sortedPersons = [...persons]
+        .sort((a, b) => {
+          return a.children[3].textContent
+            .localeCompare(b.children[3].textContent);
+        });
+
+      sortedPersons.forEach(person => tbody.append(person));
```
polosanya, line 45: You do this logic on every case. You can just do it after the switch

polosanya (changes requested):
Almost done! Please check my comments

src/scripts/main.js
JohnRust12: +

src/scripts/main.js
JohnRust12: +

src/scripts/main.js
JohnRust12: +

src/scripts/main.js
```diff
@@ -1,3 +1,46 @@
...
+      break;
+    case 'Salary':
+      sortedPeople = [...people]
+        .sort((a, b) => {
+          return a.children[3].textContent
+            .localeCompare(b.children[3].textContent);
+        });
+
+      break;
+  }
```
polosanya, line 44: Don't forget to add default case

polosanya (approved):
Good job!

PR #438, opened 2022-10-25
Develop

RusNeday:


src/scripts/main.js
```diff
@@ -1,3 +1,66 @@
...
+      <td>${list[i].name}</td>
+      <td>${list[i].position}</td>
+      <td>${list[i].age}</td>
+      <td>${list[i].salary}</td>
+    </tr>
+    `;
+  }
+};
+
+function sort(array, field) {
```
TarasHoliuk, line 57:
```suggestion
function sortEmployees(array, field) {
```
this name would be more clear, especially taking into account that you use `sort` array method below

src/scripts/main.js
```diff
@@ -1,3 +1,66 @@
...
+
+  if (e.target.innerText === 'Salary') {
+    sort(arrayPeople, e.target.innerText);
+
+    showSort(arrayPeople);
+  }
+}
+);
+
+function showSort(list) {
```
TarasHoliuk, line 42:
```suggestion
function showSorted(list) {
```

TarasHoliuk (changes requested):
Well done! I've added a few suggestions for naming

nazarmatsevych (approved):
GJ!

PR #442, opened 2022-10-29
solution

oleksii-bieliaiev:
[DEMO LINK](https://oleksii-bieliaiev.github.io/js_sort_table_DOM/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,35 @@
 'use strict';
 
-// write code here
+const tbody = document.querySelector('tbody');
+const rows = [...tbody.children];
+let sortedRows = [];
+
+document.querySelector('thead').addEventListener('click', e => {
+  sortTable(e.target.textContent);
+});
+
+function sortTable(textContent) {
+  switch (textContent) {
+    case 'Name':
+      sortedRows = rows.sort((a, b) =>
+        a.children[0].textContent.localeCompare(b.children[0].textContent));
+      break;
+    case 'Position':
+      sortedRows = rows.sort((a, b) =>
+        a.children[1].textContent.localeCompare(b.children[1].textContent));
+      break;
+    case 'Age':
+      sortedRows = rows.sort((a, b) => +a.children[2].textContent
+        - +b.children[2].textContent);
+      break;
+    case 'Salary':
+      sortedRows = rows.sort((a, b) =>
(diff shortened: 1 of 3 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,35 @@
...
+function sortTable(textContent) {
+  switch (textContent) {
+    case 'Name':
+      sortedRows = rows.sort((a, b) =>
+        a.children[0].textContent.localeCompare(b.children[0].textContent));
+      break;
+    case 'Position':
+      sortedRows = rows.sort((a, b) =>
+        a.children[1].textContent.localeCompare(b.children[1].textContent));
+      break;
```
loralevitska, line 20: avoid code duplicate. You can create function and use it (pass index you need)

src/scripts/main.js
```diff
@@ -1,3 +1,35 @@
...
+    case 'Age':
+      sortedRows = rows.sort((a, b) => +a.children[2].textContent
+        - +b.children[2].textContent);
+      break;
+    case 'Salary':
+      sortedRows = rows.sort((a, b) =>
+        +a.children[3].textContent.replace(/[^0-9.-]+/g, '')
```
loralevitska, line 27: `/[^0-9.-]+/g` - move to variable

loralevitska (changes requested):

src/scripts/main.js
oleksii-bieliaiev: Here I used a function instead of a variable

src/scripts/main.js
oleksii-bieliaiev: done

src/scripts/main.js
```diff
@@ -1,3 +1,53 @@
...
+      break;
+    case 'Salary':
+      sortTable(columnIndex.Salary);
+      break;
+    default:
+      throw new Error('Unknown step');
+  }
+});
+
+function sortTable(column) {
```
polosanya, line 32: It is a good tone to declare funtions before use, so it's better to move this logic above

polosanya (approved):
Good job! I'm approving it, but please pay attention to my comments)

src/scripts/main.js
oleksii-bieliaiev: Roger that

PR #443, opened 2022-10-29
Solution

SergIats:


src/scripts/main.js
```diff
@@ -1,3 +1,48 @@
...
+}
+
+function nameSort(list) {
+  list.sort((a, b) => (a.children[0].innerText
+    .localeCompare(b.children[0].innerText)));
+
+  mainList.append(...list);
+}
+
+menu[3].addEventListener('click', () => {
```
TarasHoliuk, line 34: use event delegation here
https://javascript.info/event-delegation

src/scripts/main.js
```diff
@@ -1,3 +1,48 @@
 'use strict';
 
-// write code here
+const employeeList = [...document.querySelectorAll('tr')].slice(1, -1);
+const mainList = document.querySelector('tbody');
+const menu = document.querySelector('thead').firstElementChild.children;
```
TarasHoliuk, line 5: with event delegation `thead` will be enough

src/scripts/main.js
```diff
@@ -1,3 +1,48 @@
 'use strict';
 
-// write code here
+const employeeList = [...document.querySelectorAll('tr')].slice(1, -1);
+const mainList = document.querySelector('tbody');
```
TarasHoliuk, line 4:
```suggestion
const tableBody = document.querySelector('tbody');
```
it is a more clear name :arrow_up:

src/scripts/main.js
```diff
@@ -1,3 +1,48 @@
...
+const menu = document.querySelector('thead').firstElementChild.children;
+
+function salarySort(list) {
```
TarasHoliuk, line 7: It's better (and necessary with event delegation) to create one "sort" function and write different behavior depending on the column (you can pass index/column inner text, etc as argument)

TarasHoliuk (changes requested):

src/scripts/main.js
SergIats: Hi, Taras! Of course, you're right. Thank you for review. I have implemented all your recommendations.

src/scripts/main.js
SergIats: done

src/scripts/main.js
SergIats: done

src/scripts/main.js
SergIats: done

vlad-demchuk (approved):

PR #452, opened 2022-11-11
solved

Deniso4ka23:
[DEMO LINK](https://Deniso4ka23.github.io/js_sort_table_DOM/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,44 @@
 'use strict';
 
-// write code here
+function getNumber(str) {
+  return +str.slice(1).replaceAll(',', '');
+}
+
+const info = document.querySelector('tbody');
+
+document.body.addEventListener('click', e => {
+  const sort = [...info.children].sort((a, b) => {
+    switch (e.target.innerText) {
+      case ('Salary'):
+        let first = getNumber(a.children[3].innerText);
+        let second = getNumber(b.children[3].innerText);
+
+        return first - second;
+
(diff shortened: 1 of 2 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,44 @@
...
+        first = a.children[1].innerText;
+        second = b.children[1].innerText;
+
+        return first.localeCompare(second);
+
+      case ('Name'):
+        first = a.children[0].innerText;
+        second = b.children[0].innerText;
+
+        return first.localeCompare(second);
```
Anastasiia-Svintsova, line 34: Look, you use the same methods here. You can combine them. Just create a variable for cell index before.  Use `e.target.cellIndex`.

Anastasiia-Svintsova (changes requested):
Good! Fix the comment

src/scripts/main.js
```diff
@@ -1,3 +1,42 @@
...
+        let first = getNumber(a.children[3].innerText);
+        let second = getNumber(b.children[3].innerText);
+
+        return first - second;
+
+      case ('Age'):
+        first = +a.children[2].innerText;
+        second = +b.children[2].innerText;
+
+        return first - second;
```
polosanya, line 26: You can also combine Salary and Age, as you compare them as numbers

polosanya (changes requested):
Almost done! Please check my and previous comments

PR #456, opened 2022-11-16
Solution

Khabailiuk:


Khabailiuk:
[DEMO LINK](https://Khabailiuk.github.io/js_sort_table_DOM/

src/scripts/main.js
```diff
@@ -1,3 +1,55 @@
...
+          rowB.cells[PositionCell].innerHTML);
+      });
+      break;
+
+    case 'Position':
+      rowsArray.sort((rowA, rowB) => {
+        return rowA.cells[PositionCell].innerHTML.localeCompare(
+          rowB.cells[PositionCell].innerHTML);
+      });
+      break;
```
vlad-demchuk, line 29: combine these checks.  for example, you don't need to do `localeCompare` twice, just combine `name` and `position` cases. make it simpler

vlad-demchuk (changes requested):

Virass (approved):

PR #458, opened 2022-11-22
Solution

KateRiaz:
[demo](https://kateriaz.github.io/js_sort_table_DOM/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,49 @@
 'use strict';
 
-// write code here
+const title = document.querySelectorAll('th');
+
+for (const titleName of title) {
+  titleName.addEventListener('click', (e) => {
+    const item = e.target;
+    const tbody = document.querySelector('tbody');
+    const rows = [...tbody.querySelectorAll('tr')];
+
+    if (item.innerText === 'Age' || item.innerText === 'Salary') {
+      let i = [...title].indexOf(item);
+
+      if (i === 6) {
+        i = 2;
+      }
+
+      if (i === 7) {
+        i = 3;
+      }
+
+      rows.sort((a, b) => Number(a.children[i].innerText.replace(/[, $]/g, ''))
+        - Number(b.children[i].innerText.replace(/[, $]/g, '')));
+
+      for (const row of rows) {
+        tbody.append(row);
+      }
+    }
+
+    if (item.innerText === 'Name' || item.innerText === 'Position') {
+      let i = [...title].indexOf(item);
+
+      if (i === 4) {
+        i = 0;
+      }
+
+      if (i === 5) {
+        i = 1;
+      }
+
+      rows.sort((a, b) => (a.children[i].innerText
+        .localeCompare(b.children[i].innerText)));
+
+      for (const row of rows) {
+        tbody.append(row);
+      }
+    }
+  });
+}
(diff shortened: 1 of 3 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,49 @@
...
+const title = document.querySelectorAll('th');
+
+for (const titleName of title) {
+  titleName.addEventListener('click', (e) => {
+    const item = e.target;
+    const tbody = document.querySelector('tbody');
+    const rows = [...tbody.querySelectorAll('tr')];
+
+    if (item.innerText === 'Age' || item.innerText === 'Salary') {
+      let i = [...title].indexOf(item);
```
Anastasiia-Svintsova, line 12: Let's use switch/case here instead of `if`

src/scripts/main.js
```diff
@@ -1,3 +1,49 @@
...
+
+      if (i === 6) {
+        i = 2;
+      }
+
+      if (i === 7) {
+        i = 3;
+      }
```
Anastasiia-Svintsova, line 20: why do you need this manipulation with `i`?

Anastasiia-Svintsova (changes requested):

src/scripts/main.js
KateRiaz: to make footer to sort rows

src/scripts/main.js
```diff
@@ -1,3 +1,34 @@
...
+const title = document.querySelectorAll('th');
+
+for (const titleName of title) {
+  titleName.addEventListener('click', (e) => {
+    const item = e.target;
+    const tbody = document.querySelector('tbody');
+    const rows = [...tbody.querySelectorAll('tr')];
+
+    switch (item.innerText) {
+      case 'Age' || 'Salary':
```
loralevitska, line 12:
```suggestion
      case 'Age':
      case 'Salary':
```

src/scripts/main.js
```diff
@@ -1,3 +1,34 @@
...
+        const i = [...title].indexOf(item);
+
+        rows.sort((a, b) =>
+          Number(a.children[i].innerText.replace(/[, $]/g, ''))
+          - Number(b.children[i].innerText.replace(/[, $]/g, '')));
+
+        for (const row of rows) {
+          tbody.append(row);
+        }
```
loralevitska, line 21: DRY. There is a duplicate code in lines 28-30

do this operation outside `switch case`

src/scripts/main.js
```diff
@@ -1,3 +1,34 @@
...
+        break;
+
+      case 'Name' || 'Position':
```
loralevitska, line 24:
```suggestion
      case 'Name':
      case 'Position':
```

loralevitska (changes requested):
almost done)

src/scripts/main.js
```diff
@@ -1,3 +1,31 @@
...
+      case 'Salary':
+      case 'Age':
+        rows.sort((a, b) =>
+          Number(a.children[i].innerText.replace(/[, $]/g, ''))
+          - Number(b.children[i].innerText.replace(/[, $]/g, '')));
+        break;
+
+      case 'Name':
+      case 'Position':
+        rows.sort((a, b) => (a.children[i].innerText
```
polosanya, line 22: You could also make it just a default case

polosanya (approved):
Good job! I'm approving it, but please check my comment)

PR #460, opened 2022-11-27
vol_1

RudBeee:
- [DEMO LINK](https://RudBeee.github.io/js_sort_table_DOM/)


tetiana-ivanchenko (changes requested):
if i sort 1 column, all is ok, if I sort second column - it sorted by desc, instead of asc

TarasHoliuk (approved):

PR #465, opened 2022-12-05
Solution

ifuride:
[DEMO LINK](https://ifuride.github.io/js_sort_table_DOM/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,34 @@
 'use strict';
 
-// write code here
+const tableHead = document.querySelector('thead');
+const tableBody = document.querySelector('tbody');
+const rows = [...tableBody.rows];
+
+tableHead.addEventListener('click', (handler) => {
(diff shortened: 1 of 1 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,34 @@
...
+
+  switch (header) {
+    case 'Name':
+      rows.sort((a, b) =>
+        (a.children[0].innerText).localeCompare(b.children[0].innerText));
+      break;
+
+    case 'Position':
+      rows.sort((a, b) =>
+        (a.children[1].innerText).localeCompare(b.children[1].innerText));
```
tetiana-ivanchenko, line 18: combine it

tetiana-ivanchenko (changes requested):

src/scripts/main.js
ifuride: Could you please elaborate? I don't think I understand how to do it. Different indexes are used for Name and Position.

PR #470, opened 2022-12-08
Sort table

maksym-kryvolap:
[DEMO LINK](https://maksym-kryvolap.github.io/js_sort_table_DOM/)


vlad-demchuk (changes requested):
let's simplify it

create a common function for sorting to avoid repetitions

AlisaVlasova (approved):

PR #473, opened 2022-12-09
Solution

Ivanenko1402:
[DEMO LINK](https://Ivanenko1402.github.io/js_sort_table_DOM/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,90 @@
 'use strict';
 
-// write code here
+const tHead = document.querySelector('thead');
+const tBody = document.querySelector('tbody');
+const rows = [...tBody.children];
+let reverse = false;
+
+function handleEvent(e) {
+  let newList = [...rows];
+
+  if (reverse) {
+    reverse = false;
+
+    switch (e.target.textContent) {
+      case 'Name':
+        newList = [...rows].sort(
+          (a, b) =>
+            b.children[0].textContent.localeCompare(a.children[0].textContent)
+        );
+        break;
+
+      case 'Position':
+        newList = [...rows].sort(
+          (a, b) =>
+            b.children[1].textContent.localeCompare(a.children[1].textContent)
+        );
+        break;
+
+      case 'Age':
+        newList = [...rows].sort(
+          (a, b) =>
+            +b.children[2].textContent - +a.children[2].textContent
+        );
+        break;
+
+      case 'Salary':
+        newList = [...rows].sort(
+          (a, b) =>
+            +b.children[3].textContent.slice(1).split(',').join('')
+              - +a.children[3].textContent.slice(1).split(',').join('')
+        );
+        break;
+
+      default:
+        break;
+    }
+  } else {
+    reverse = true;
+
+    switch (e.target.textContent) {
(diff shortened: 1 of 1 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,90 @@
 'use strict';
 
-// write code here
+const tHead = document.querySelector('thead');
+const tBody = document.querySelector('tbody');
+const rows = [...tBody.children];
+let reverse = false;
+
+function handleEvent(e) {
```
polosanya, line 8:
```suggestion
function handleEvent(event) {
```
Let's use full, descriptive names

src/scripts/main.js
```diff
@@ -1,3 +1,90 @@
...
+  let newList = [...rows];
+
+  if (reverse) {
+    reverse = false;
```
polosanya, line 12: I think it's better to move this check inside cases.
For now, you have two different switch...cases, but everything can be done with only one.
Let's try to optimize your code

src/scripts/main.js
```diff
@@ -1,3 +1,90 @@
...
+        newList = [...rows].sort(
+          (a, b) =>
+            b.children[0].textContent.localeCompare(a.children[0].textContent)
+        );
+        break;
+
+      case 'Position':
+        newList = [...rows].sort(
+          (a, b) =>
+            b.children[1].textContent.localeCompare(a.children[1].textContent)
```
polosanya, line 25: In both cases you do almost the same. Think about combining similar cases. 
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/switch#taking_advantage_of_fall-through

src/scripts/main.js
```diff
@@ -1,3 +1,90 @@
...
+            +b.children[2].textContent - +a.children[2].textContent
+        );
+        break;
+
+      case 'Salary':
+        newList = [...rows].sort(
+          (a, b) =>
+            +b.children[3].textContent.slice(1).split(',').join('')
+              - +a.children[3].textContent.slice(1).split(',').join('')
+        );
```
polosanya, line 41: Same with age and salary. Also, please create variables for duplicated values or/and create helping functions for duplicated actions to make your code cleaner

polosanya (changes requested):
Good job! Let's improve your code a bit

polosanya (changes requested):
Do you really need to have two same switch...cases? It looks complicated and not optimized.
Just move logic from sortedBy function to handleEvent and try to combine them.

src/scripts/main.js
```diff
@@ -1,3 +1,42 @@
...
+  const header = [...document.querySelectorAll('th')];
+  const num = header.findIndex(
+    item => item.textContent === e.target.textContent
+  );
+
+  if (value === e.target.textContent) {
+    reverse = !reverse;
+  } else {
+    reverse = false;
+  }
```
Anastasiia-Svintsova, line 19: You also could use a ternary operator here 
```suggestion
  reverse = value === e.target.textContent ? !reverse : false;
```

Anastasiia-Svintsova (approved):

PR #480, opened 2022-12-13
code for sorting table was added

Olena-Shevchenko:
- [DEMO LINK](https://Olena-Shevchenko.github.io/js_sort_table_DOM/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,39 @@
 'use strict';
 
-// write code here
+const table = document.querySelector('table');
+const tBody = document.querySelector('tbody');
+const tr = tBody.querySelectorAll('tr');
+
+table.addEventListener('click', (e) => {
+  const header = e.target.closest('th');
+
+  if (header) {
+    const trArr = [...tr];
+
+    if (e.target.textContent === 'Name') {
+      trArr.sort((a, b) =>
+        a.children[0].textContent
+          .localeCompare(b.children[0].textContent));
+    } else if (e.target.textContent === 'Position') {
+      trArr.sort((a, b) =>
+        a.children[1].textContent
+          .localeCompare(b.children[1].textContent));
+    } else if (e.target.textContent === 'Age') {
+      trArr.sort((a, b) => +a.children[2].textContent
+        - +b.children[2].textContent);
+    } else if (e.target.textContent === 'Salary') {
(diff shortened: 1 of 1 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,39 @@
...
+  const header = e.target.closest('th');
+
+  if (header) {
+    const trArr = [...tr];
+
+    if (e.target.textContent === 'Name') {
+      trArr.sort((a, b) =>
+        a.children[0].textContent
+          .localeCompare(b.children[0].textContent));
+    } else if (e.target.textContent === 'Position') {
```
polosanya, line 17: You can use [switch...case](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/switch) instead of if...else.
It will make your code cleaner and more understandable

src/scripts/main.js
```diff
@@ -1,3 +1,39 @@
 'use strict';
 
-// write code here
+const table = document.querySelector('table');
+const tBody = document.querySelector('tbody');
+const tr = tBody.querySelectorAll('tr');
```
polosanya, line 5: Let's try to provide more meaningful names, not just copy selectors

src/scripts/main.js
```diff
@@ -1,3 +1,39 @@
...
+      });
+    }
+
+    for (const oneTr of trArr) {
+      tBody.append(oneTr);
+    }
+  }
+});
+
+function convertSalary(salary) {
```
polosanya, line 37: It is a good tone to declare function before use, so please move it to the top

polosanya (changes requested):
Almost perfect! Let's provide some changes

src/scripts/main.js
```diff
@@ -1,3 +1,44 @@
...
+        rows.sort((a, b) =>
+          a.children[0].textContent
+            .localeCompare(b.children[0].textContent));
+        break;
+
+      case 'Position':
+        rows.sort((a, b) =>
+          a.children[1].textContent
+            .localeCompare(b.children[1].textContent));
+        break;
```
loralevitska, line 26: Look at the duplicate code you have here. Move everything that is duplicated to the helper function

loralevitska (changes requested):

agniya-i (approved):

PR #481, opened 2022-12-13
Solution

Vlad-Fedorishchev:
Demo Link  -  https://Vlad-Fedorishchev.github.io/js_sort_table_DOM/

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -10,39 +10,36 @@ for (const element of navListElements) {
       return item.textContent;
     };
 
-    if (action.target === navListElements[0]) {
-      const sortedNames = tbodyChildrens.sort(
-        (a, b) => getTextContent(a.children[0]).localeCompare(
-          getTextContent(b.children[0])));
+    let sortedRows = null;
 
-      tbody.append(...sortedNames);
+    switch (action.target) {
+      case navListElements[0]:
+        sortedRows = tbodyChildrens.sort(
+          (a, b) => getTextContent(a.children[0]).localeCompare(
+            getTextContent(b.children[0])));
+        break;
+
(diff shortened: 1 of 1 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,48 @@
...
+const tbodyChildrens = [...document.querySelector('tbody').children];
+const navListElements = [...document.body.querySelector('tr').children];
+
+for (const element of navListElements) {
+  element.addEventListener('click', (action) => {
+    const getTextContent = (item) => {
+      return item.textContent;
+    };
+
+    if (action.target === navListElements[0]) {
```
polosanya, line 13: It is better to use [switch...case](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/switch) if you know all the possible variants of action.taget.
It will make your code cleaner

src/scripts/main.js
```diff
@@ -1,3 +1,48 @@
...
+    if (action.target === navListElements[3]) {
+      const makeNumber = (string) => {
+        return +string.slice(1).split(',').join('');
+      };
+
+      const sortedSalary = tbodyChildrens.sort(
+        (a, b) => (makeNumber(a.children[3].textContent))
+            - (makeNumber(b.children[3].textContent)));
+
+      tbody.append(...sortedSalary);
```
polosanya, line 45: In all cases you create some variable and append it to tbody. 
Just create a variable in the upper scope, set a value inside cases and append it after all checks.

polosanya (changes requested):
Almost done! Let's improve your work a bit

loralevitska (approved):
good job!

PR #488, opened 2022-12-21
solution

Deniskakaka:


src/scripts/main.js
```diff
@@ -1,3 +1,50 @@
...
+
+  people.push({
+    name: namePerson.innerText,
+    position: position.innerText,
+    age: age.innerText,
+    salary: salary.innerText,
+  });
+}
+
+const functionFoSort = (string) => {
```
OlEzhA0, line 21:
```suggestion
const sortFunction = (sortKey) => {
```

src/scripts/main.js
```diff
@@ -1,3 +1,50 @@
...
+};
+
+for (let i = 0; i < body.children.length; i++) {
+  const [ namePerson, position, age, salary ] = body.children[i].children;
+
+  people.push({
+    name: namePerson.innerText,
+    position: position.innerText,
+    age: age.innerText,
+    salary: salary.innerText,
```
OlEzhA0, line 17: you can convert salary and age to number for your sortable data at once

src/scripts/main.js
```diff
@@ -1,3 +1,50 @@
...
+const functionFoSort = (string) => {
+  if (string === 'Name' || string === 'Position') {
+    return (a, b) => a[string.toLowerCase()]
+      .localeCompare(b[[string.toLowerCase()]]);
+  }
+
+  if (string === 'Age' || string === 'Salary') {
+    return (a, b) =>
+      getNumber(a[string.toLowerCase()]) - getNumber(b[string.toLowerCase()]);
+  }
```
OlEzhA0, line 30: I see you can return like this (example)
```suggestion
  return (a, b) => {
    const firstElement = a[string.toLowerCase()]
    const secondElement = b[string.toLowerCase()]
    
    if (string === 'Name') {
      return firstElement.localeCompare(secondElement)
    }
    
    if (string === 'Age) {
      return firstElement - secondElement
    }
  }
```

OlEzhA0: Just to follow DRY principle (Don't Repeat Yourself)

src/scripts/main.js
```diff
@@ -1,3 +1,50 @@
...
+};
+
+const sortBy = document.querySelectorAll('th');
+
+sortBy.forEach(el => {
+  el.addEventListener('click', (e) => {
+    body.innerHTML = '';
+    people.sort(functionFoSort(e.target.innerText));
```
OlEzhA0, line 38: you can lower your key for sort right here and don't do it each time in function
```suggestion
    people.sort(functionFoSort(e.target.innerText.toLowerCase()));
```

src/scripts/main.js
```diff
@@ -1,3 +1,50 @@
...
+
+    people.forEach(q => {
```
OlEzhA0, line 40: what is a `q`?

OlEzhA0 (changes requested):
hey! please check my comments

src/scripts/main.js
Viktor-Kost: is not fixed

src/scripts/main.js
Viktor-Kost: This conversation is commented at fe_chat by mentor

Viktor-Kost (changes requested):
Hello! 
Pay attention to unresolved conversations. If you need any help - ask at fe_chat

witflash (approved):
👍🏼

PR #517, opened 2023-01-18
first attempt

IhorChaikovskyi:
[DEMO LINK](https://IhorChaikovskyi.github.io/js_sort_table_DOM/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,101 @@
 'use strict';
 
-// write code here
+const head = document.querySelector('thead');
+const body = document.querySelector('tbody');
+const bodychildren = [...body.children];
+
+function transformationNumber(element) {
+  const array = element.children[3].textContent.slice(1).split(',');
+
+  return +(array[0] + array[1]);
+}
+
+head.addEventListener('click', even => {
+  if (even.target.textContent === 'Age') {
+    bodychildren.sort((one, two) => {
+      return +one.children[2].textContent - +two.children[2].textContent;
+    });
+
+    if (even.target.dataset.clicked === undefined) {
+      for (const child of bodychildren) {
+        body.append(child);
+      }
+      even.target.dataset.clicked = true;
+
+      return;
+    }
+
+    if (even.target.dataset.clicked === 'true') {
+      for (const child of bodychildren) {
+        body.prepend(child);
+      }
+      even.target.removeAttribute('data-clicked');
+    }
+  }
+
+  if (even.target.textContent === 'Position') {
+    bodychildren.sort((a, b) => {
+      return a.children[1].textContent.localeCompare(b.children[1].textContent);
+    });
+
+    if (even.target.dataset.clicked === undefined) {
+      for (const child of bodychildren) {
+        body.append(child);
+      }
+      even.target.dataset.clicked = true;
+
+      return;
+    }
+
+    if (even.target.dataset.clicked === 'true') {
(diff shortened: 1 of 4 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,101 @@
...
+const bodychildren = [...body.children];
+
+function transformationNumber(element) {
+  const array = element.children[3].textContent.slice(1).split(',');
+
+  return +(array[0] + array[1]);
+}
+
+head.addEventListener('click', even => {
+  if (even.target.textContent === 'Age') {
```
tetiana-ivanchenko, line 14: use switch/case

src/scripts/main.js
```diff
@@ -1,3 +1,101 @@
...
+
+      return;
+    }
+
+    if (even.target.dataset.clicked === 'true') {
+      for (const child of bodychildren) {
+        body.prepend(child);
+      }
+      even.target.removeAttribute('data-clicked');
+    }
```
tetiana-ivanchenko, line 33: why did you need this code?

tetiana-ivanchenko (changes requested):

src/scripts/main.js
IhorChaikovskyi: These conditions add elements to the column. The first condition adds in ascending order, the other in descending order.

src/scripts/main.js
```diff
@@ -1,3 +1,59 @@
...
+
+function transformationNumber(element) {
+  const array = element.children[3].textContent.slice(1).split(',');
+
+  return +(array[0] + array[1]);
+}
+
+head.addEventListener('click', even => {
+  switch (true) {
+    case even.target.textContent === 'Age' :
```
tetiana-ivanchenko, line 15:
```suggestion
  switch (even.target.textContent) {
    case 'Age' :
```

src/scripts/main.js
```diff
@@ -1,3 +1,59 @@
...
+          b.children[0].textContent
+        );
+      });
+      break;
+
+    case even.target.textContent === 'Salary' :
+      bodychildren.sort((a, b) => {
+        return transformationNumber(a) - transformationNumber(b);
+      });
+      break;
```
tetiana-ivanchenko, line 41: create the function which takes an index, and sort by index - in this way you don't need switch/case

tetiana-ivanchenko (changes requested):

src/scripts/main.js
```diff
@@ -1,3 +1,44 @@
 'use strict';
 
-// write code here
+const head = document.querySelector('thead');
+const body = document.querySelector('tbody');
```
TarasHoliuk, line 4:
```suggestion
const tableBody = document.querySelector('tbody');
```
`body` could be the tag `<body>`

the same with `head`

src/scripts/main.js
```diff
@@ -1,3 +1,44 @@
...
+const bodychildren = [...body.children];
```
TarasHoliuk, line 5: `tableRows` would be more clear name

src/scripts/main.js
```diff
@@ -1,3 +1,44 @@
...
+
+function transformationNumber(element) {
```
TarasHoliuk, line 7:
```suggestion
function transformToNumber(element) {
```
it's better to name functions in this way :arrow_up: (when it contains **verb** that explains what the function is doing)

TarasHoliuk (approved):
:+1:

PR #521, opened 2023-01-22
Sort table

katerinalex:
[DEMO LINK](https://katerinalex.github.io/js_sort_table_DOM/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,77 @@
 'use strict';
 
-// write code here
+const table = document.querySelector('table');
+const thead = table.querySelector('thead');
+const tbody = table.querySelector('tbody');
+const headers = thead.querySelector('tr');
+const tr = tbody.querySelectorAll('tr');
+let sorted = [];
+
+function sortStrings(data, key) {
+  const dataTable = [...data];
+
+  const newDataTable = dataTable.sort((a, b) => {
+    const stringA = a.children[key].textContent;
+    const stringB = b.children[key].textContent;
+
+    if (stringA > stringB) {
+      return 1;
+    } else if (stringA === stringB) {
+      return 0;
+    } else {
+      return -1;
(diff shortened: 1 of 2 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,77 @@
...
+    const stringA = a.children[key].textContent;
+    const stringB = b.children[key].textContent;
+
+    if (stringA > stringB) {
+      return 1;
+    } else if (stringA === stringB) {
+      return 0;
+    } else {
+      return -1;
+    }
```
polosanya, line 23:
```suggestion
    if (stringA > stringB) {
      return 1;
    }
    
    if (stringA === stringB) {
      return 0;
    }
    
    return -1;
```
No need for else. You code will not go further after return anyway

polosanya (changes requested):
Your link doesn't work(
Please make sure that you've deployed everything
image

src/scripts/main.js
Anastasiia-Svintsova: Not fixed 
image

src/scripts/main.js
```diff
@@ -1,3 +1,77 @@
...
+}
+
+function sortNumbers(data, key) {
+  const dataTable = [...data];
+
+  const newDataTable = dataTable.sort((a, b) => {
+    const stringA = a.children[key].textContent;
+    const stringB = b.children[key].textContent;
+    const numA = parseInt(stringA.split(',').join('').split('$').join(''));
+    const numB = parseInt(stringB.split(',').join('').split('$').join(''));
```
Anastasiia-Svintsova, line 36: You do almost the same in both cases. Create a method for it

Anastasiia-Svintsova (changes requested):

Anastasiia-Svintsova (approved):

PR #523, opened 2023-01-23
DOM / sort table

valntyn:
[DEMO LINK](https://valntyn.github.io/js_sort_table_DOM/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,42 @@
 'use strict';
 
-// write code here
+const menu = document.querySelector('thead');
+const body = document.querySelector('tbody');
+const list = body.querySelectorAll('tr');
+
+menu.addEventListener('click', e => {
+  const item = e.target.closest('th');
+  const index = e.target.cellIndex;
+
+  if (!item || !menu.contains(item)) {
+    return;
+  }
+
+  const sortedCells = [...list]
+    .sort((a, b) => {
+      const aText = a.children[index].innerText;
+      const bText = b.children[index].innerText;
+
+      switch (index) {
+        case 0:
+        case 1:
+          return aText.localeCompare(bText);
+
+        case 2:
+          return +aText - +bText;
+
+        case 3:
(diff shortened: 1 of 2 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,42 @@
...
+    return;
+  }
+
+  const sortedCells = [...list]
+    .sort((a, b) => {
+      const aText = a.children[index].innerText;
+      const bText = b.children[index].innerText;
+
+      switch (index) {
+        case 0:
```
OlEzhA0, line 21: Would be better to sort by labels, not index
if you have column reordering it's hard to fix this function for sorting

src/scripts/main.js
```diff
@@ -1,3 +1,42 @@
...
+          return +aText - +bText;
+
+        case 3:
+          const aNum = +(aText.replace(/\W/g, ''));
+          const bNum = +(bText.replace(/\W/g, ''));
+
+          return aNum - bNum;
+
+        default:
+          return 0;
```
OlEzhA0, line 35: why do we need return 0 here?
just 
```
default:
  break;
```

src/scripts/main.js
```diff
@@ -1,3 +1,42 @@
...
+      }
+    });
+
+  sortedCells.forEach(el => {
+    return body.append(el);
+  });
```
OlEzhA0, line 41:
```suggestion
  sortedCells.forEach(el => body.append(el));
```

or even

```suggestion
  sortedCells.forEach(body.append);
```

OlEzhA0 (changes requested):

src/scripts/main.js
valntyn: thank you, has been fixed

src/scripts/main.js
valntyn: thank you, has been fixed

OlEzhA0 (approved):

PR #525, opened 2023-01-25
Finish task

U-Ra-All:
- [DEMO LINK](https://U-Ra-All.github.io/js_sort_table_DOM/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,15 +1,15 @@
-"use strict";
+'use strict';
 
-const employees = document.querySelector("table");
+const employees = document.querySelector('table');
 
 sortTable(employees);
 
 function sortTable(table) {
-  document.querySelectorAll("th").forEach((th) =>
-    th.addEventListener("click", () => {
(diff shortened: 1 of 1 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,43 @@
...
+      && !isNaN(aCellValue)
+      && !isNaN(bCellValue)
+    ) {
+      return aCellValue - bCellValue;
+    }
+
+    return aCellValue[0] === '$'
+      ? parseInt(aCellValue.slice(1)) - parseInt(bCellValue.slice(1))
+      : aCellValue.toString().localeCompare(bCellValue);
+  };
```
OlEzhA0, line 38: would be better to add switch case, and add sort by column label like
```
switch (key):
  case 'Position':
  case 'Name':
    return a - b;

  case 'Salary':
    return parseInt(a.slice(1)) - parseInt(b.slice(1));
```

etc

OlEzhA0 (changes requested):

vlad-demchuk (approved):

PR #528, opened 2023-01-30
sort table

anastasia-radchencko:
https://anastasia-radchencko.github.io/js_sort_table_DOM/

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,43 @@
 'use strict';
 
 // write code here
+const table = document.querySelector('table');
+const tbody = document.querySelector('tbody');
+const row = [...tbody.children];
+
+table.addEventListener('click', (e) => {
+  const th = e.target.closest('th');
+
+  if (!th || !table.contains(th)) {
+    return;
+  }
+
+  sortTable(th.cellIndex, th.innerHTML);
+});
+
+function sortTable(index, text) {
+  switch (text) {
+    case 'Name':
+    case 'Position':
+      row.sort((a, b) =>
+        a.cells[index].innerHTML.localeCompare(b.cells[index].innerHTML));
+      break;
+
+    case 'Age':
+      row.sort((a, b) =>
+        a.cells[index].innerHTML - b.cells[index].innerHTML);
+      break;
+
+    case 'Salary':
+      row.sort((a, b) =>
+        normalNumber(a.cells[index].innerHTML)
+          - normalNumber(b.cells[index].innerHTML));
+      break;
+  }
+
+  tbody.append(...row);
+}
+
+function normalNumber(number) {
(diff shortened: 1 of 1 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,43 @@
 'use strict';
 
 // write code here
+const table = document.querySelector('table');
+const tbody = document.querySelector('tbody');
+const row = [...tbody.children];
```
DarkMistyRoom, line 6: you can do the same with `table.rows`

src/scripts/main.js
```diff
@@ -1,3 +1,43 @@
...
+      row.sort((a, b) =>
+        normalNumber(a.cells[index].innerHTML)
+          - normalNumber(b.cells[index].innerHTML));
+      break;
+  }
+
+  tbody.append(...row);
+}
+
+function normalNumber(number) {
```
DarkMistyRoom, line 41:
```suggestion
function normalizeNumber(number) {
```

src/scripts/main.js
```diff
@@ -1,3 +1,43 @@
...
+      row.sort((a, b) =>
+        a.cells[index].innerHTML - b.cells[index].innerHTML);
+      break;
+
+    case 'Salary':
+      row.sort((a, b) =>
+        normalNumber(a.cells[index].innerHTML)
+          - normalNumber(b.cells[index].innerHTML));
+      break;
+  }
```
DarkMistyRoom, line 36: add default value

DarkMistyRoom (changes requested):
Good job! but always handle the default case in switch construction

src/scripts/main.js
```diff
@@ -1,3 +1,47 @@
 'use strict';
 
 // write code here
+const table = document.querySelector('table');
+const tbody = document.querySelector('tbody');
+const tr = tbody.querySelectorAll('tr');
+const row = [...tr];
```
Volodymyr-Melnyk-Fron10der, line 7: I think Misha did mean like:
```suggestion
const table = document.querySelector('table');
const tbody = document.querySelector('tbody');
const row = [...table.rows];
```

src/scripts/main.js
```diff
@@ -1,3 +1,47 @@
...
+
+table.addEventListener('click', (e) => {
+  const th = e.target.closest('th');
+
+  if (!th || !table.contains(th)) {
+    return;
+  }
+
+  sortTable(th.cellIndex, th.innerHTML);
```
Volodymyr-Melnyk-Fron10der, line 16: It looks easier for me)
```suggestion
  if (th && table.contains(th)) {
     sortTable(th.cellIndex, th.innerHTML);
  } 
```

Volodymyr-Melnyk-Fron10der (approved):
Great! I left some recommendation only

PR #549, opened 2023-02-16
added

vovakov123:
- [DEMO LINK](https://vovakov123.github.io/js_sort_table_DOM/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,44 @@
 'use strict';
 
-// write code here
+const thead = document.querySelector('thead');
+const tbody = document.querySelector('tbody');
+const tr = tbody.rows;
+
+thead.addEventListener('click', (e) => {
+  if (e.target.textContent === 'Name') {
+    const sortNames = function(arr) {
+      return [...arr].sort((a, b) =>
+        a.children[0].textContent < b.children[0].textContent ? -1 : 1);
+    };
+
+    tbody.append(...sortNames(tr));
+  }
+
+  if (e.target.textContent === 'Position') {
+    const sortNames = function(arr) {
+      return [...arr].sort((a, b) =>
+        a.children[1].textContent < b.children[1].textContent ? -1 : 1);
+    };
+
+    tbody.append(...sortNames(tr));
+  }
+
+  if (e.target.textContent === 'Age') {
+    const sortNames = function(arr) {
+      return [...arr].sort((a, b) =>
+        a.children[2].textContent < b.children[2].textContent ? -1 : 1);
+    };
+
+    tbody.append(...sortNames(tr));
+  }
+
+  if (e.target.textContent === 'Salary') {
+    const sortNames = function(arr) {
+      return [...arr].sort((a, b) =>
+        +a.children[3].textContent.slice(1).split(',').join('')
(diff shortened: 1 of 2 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,44 @@
 'use strict';
 
-// write code here
+const thead = document.querySelector('thead');
+const tbody = document.querySelector('tbody');
+const tr = tbody.rows;
+
+thead.addEventListener('click', (e) => {
+  if (e.target.textContent === 'Name') {
```
Anastasiia-Svintsova, line 8: Use switch/case instead of if statements

src/scripts/main.js
```diff
@@ -1,3 +1,44 @@
...
+    };
+
+    tbody.append(...sortNames(tr));
+  }
+
+  if (e.target.textContent === 'Age') {
+    const sortNames = function(arr) {
+      return [...arr].sort((a, b) =>
+        a.children[2].textContent < b.children[2].textContent ? -1 : 1);
+    };
```
Anastasiia-Svintsova, line 30: You do the same actions for text columns. So just combine these cases

Anastasiia-Svintsova (changes requested):

src/scripts/main.js
```diff
@@ -1,3 +1,48 @@
...
+      break;
+
+    case 'Position':
+      sortNames = function(arr) {
+        return [...arr].sort((a, b) =>
+          a.children[1].textContent < b.children[1].textContent ? -1 : 1);
+      };
+
+      tbody.append(...sortNames(tr));
+      break;
```
tetiana-ivanchenko, line 26: combine it

tetiana-ivanchenko (changes requested):

src/scripts/main.js
vovakov123: how to combine if I have different array elements?  
children[0] and children[1]
help please

anstsot: you can create 2 different functions for sorting text and number values: `sortText` and `sortNumbers`.
and then you can just reuse them.
let's consider the example:
```
const sortText = (data, childIdx) => data
  .sort((a, b) => a.children[childIdx].textContent < b.children[childIdx].textContent ? -1 : 1);
```
then you will just reuse the function as `sortText([...arr], 0)`

anstsot (changes requested):
Looks good, let's make the last change and combine the sorting functions

src/scripts/main.js
witflash: Hi!
You can create common function like this:
```
const sortByIndex = (items, index) => [...items].sort(a, b) => 
  a.children[index].textContent < b.children[index].textContent ? -1 : 1;
```
And use it for all cases:
```
case 'Name':
  tbody.append(sortByIndex(tr, 0));
  break;

case 'Position': 
  tbody.append(sortByIndex(tr, 0));
  break;

...
```

witflash (changes requested):
Check my comment, please.

src/scripts/main.js
vovakov123: > Check my comment, please.

Thank

witflash (approved):
Much better!
Well done 👍

PR #554, opened 2023-02-19
first commit

haduigon:
https://haduigon.github.io/js_sort_table_DOM/

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,131 @@
 'use strict';
 
 // write code here
+
+const table = document.getElementsByTagName('table');
+
+const people = table[0].children[1].children;
+
+table[0].children[0].children[0].children[0].addEventListener('click',
+  () => sortByName());
+
+table[0].children[0].children[0].children[1].addEventListener('click',
+  () => sortByPosition());
+
+table[0].children[0].children[0].children[2].addEventListener('click',
+  () => sortByAge());
+
+table[0].children[0].children[0].children[3].addEventListener('click',
+  () => sortBySalary());
+
+function sortByName() {
+  let switching = true;
+
+  let shouldSwitch;
+
+  let x, y, i;
+
+  while (switching) {
+    switching = false;
+
+    for (i = 0; i < (people.length - 1); i++) {
+      shouldSwitch = false;
+      x = people[i].children[0].innerText.toLowerCase();
+      y = people[i + 1].children[0].innerText.toLowerCase();
+
+      if (x.localeCompare(y) === 1) {
+        shouldSwitch = true;
+        break;
+      }
+    }
+
+    if (shouldSwitch) {
+      people[i].parentNode.insertBefore(people[i + 1], people[i]);
+      switching = true;
+    }
+  }
+}
+
+function sortByPosition() {
+  let switching = true;
+
+  let shouldSwitch;
+
+  let x, y, i;
+
+  while (switching) {
+    switching = false;
+
+    for (i = 0; i < (people.length - 1); i++) {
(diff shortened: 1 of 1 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,131 @@
...
+    }
+
+    if (shouldSwitch) {
+      people[i].parentNode.insertBefore(people[i + 1], people[i]);
+      switching = true;
+    }
+  }
+}
+
+function sortByAge() {
```
TarasHoliuk, line 105: You are duplicating a lot of code in these functions
Instead of four functions create one that accepts the index of the column as an argument and sorts it accordingly, also, you may join some cases. (`switch case` block will be useful)

Also, you may transform the HTML collection to Array in this way: `[...people]` and use built-in sort function

src/scripts/main.js
```diff
@@ -1,3 +1,131 @@
 'use strict';
 
 // write code here
+
+const table = document.getElementsByTagName('table');
```
TarasHoliuk, line 5:
```suggestion
const table = document.querySelector('table');
```
There is only one table

src/scripts/main.js
```diff
@@ -1,3 +1,131 @@
...
+
+const people = table[0].children[1].children;
+
+table[0].children[0].children[0].children[0].addEventListener('click',
```
TarasHoliuk, line 9: try to avoid so many nested children call (it's not clear what is it)
anyway better to use [eventDelagation](https://javascript.info/event-delegation) in this case

TarasHoliuk (changes requested):

src/scripts/main.js
haduigon: have rewritten the logic ) now it is only one main function and one call)

src/scripts/main.js
Viktor-Kost: is not fixed

src/scripts/main.js
```diff
@@ -1,3 +1,79 @@
...
+  let newIndex = index;
+
+  if (index === 1) {
+    newIndex = 0;
+  }
+
+  let clearData;
+
+  switch (newIndex) {
+    case 0:
```
TarasHoliuk, line 25:
```suggestion
    case 0:
    case 1:
```

instead of this code:
```
if (index === 1) {
    newIndex = 0;
  }
```

src/scripts/main.js
```diff
@@ -1,3 +1,79 @@
 'use strict';
 
 // write code here
+
+const table = document.getElementsByTagName('table');
+
+const people = table[0].children[1].children;
+
+const headers = [...table[0].children[0].children[0].children];
```
TarasHoliuk, line 9: again, a lot of `children[0]` doesn't look clear
you may query all `th`, or query `thead` and get their children

also, it's better to save some values in variables
for example query `tbody`, save it in a variable and use it when you need any first level children

TarasHoliuk (changes requested):

tetiana-ivanchenko (changes requested):
fix prev comments

haduigon:
> fix prev comments

did it

src/scripts/main.js
```diff
@@ -1,3 +1,81 @@
...
+  let clearData;
+
+  switch (index) {
+    case 0:
+      clearData = data.toLowerCase();
+      break;
+
+    case 1:
+      clearData = data.toLowerCase();
+      break;
```
sTorba24, line 29: you do the same action for 2 cases, so do as suggested above

sTorba24 (approved):
Good work! But check comments more carefully :)

PR #555, opened 2023-02-19
Solution develop

NazariiAlieksieiev:
[DEMO LINK](https://NazariiAlieksieiev.github.io/js_sort_table_DOM/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,21 @@
 'use strict';
 
-// write code here
+const table = document.querySelector('table');
+const tbody = document.querySelector('tbody');
+const rows = [...tbody.querySelectorAll('tr')];
+
+table.addEventListener('click', e => {
+  if (e.target.tagName === 'TH') {
+    const headerIndex = [...e.target.parentNode.children].indexOf(e.target);
+
+    rows.sort((a, b) => {
+      const aData = a.children[headerIndex].textContent;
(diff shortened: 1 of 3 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,21 @@
...
+    rows.sort((a, b) => {
+      const aData = a.children[headerIndex].textContent;
+      const bData = b.children[headerIndex].textContent;
+
+      return aData.localeCompare(bData, undefined, {
+        numeric: true, sensitivity: 'base',
+      });
+    });
+    rows.forEach((row) => tbody.appendChild(row));
+  }
```
VitaliyBondarenko1982, line 20:
```suggestion
  if (e.target.tagName !== 'TH')  {
    return;
  }
  
    const headerIndex = [...e.target.parentNode.children].indexOf(e.target);

    rows.sort((a, b) => {
      const aData = a.children[headerIndex].textContent;
      const bData = b.children[headerIndex].textContent;

      return aData.localeCompare(bData, undefined, {
        numeric: true, sensitivity: 'base',
      });
    });
    rows.forEach((row) => tbody.appendChild(row));
```

VitaliyBondarenko1982 (approved):
Good job.

It is just suggestion, that exist approach called early return after if. Maybe in this way function will be more readable.

PR #557, opened 2023-02-22
solution

Chupiak:
[DEMO LINK](https://chupiak.github.io/js_sort_table_DOM/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,33 @@
 'use strict';
 
-// write code here
+const headElements = document
+  .querySelector('thead')
+  .querySelectorAll('th');
+
+for (const item of headElements) {
+  item.addEventListener('click', () => {
+    const table
+      = [...document.querySelector('tbody')
+        .querySelectorAll('tr')]
+        .sort(compareValues);
+
+    function compareValues(aInput, bInput) {
+      let a = aInput.children[item.cellIndex].textContent;
+      let b = bInput.children[item.cellIndex].textContent;
+
+      if (convert(a) > 0) {
+        a = (convert(a));
+        b = (convert(b));
+      }
+
+      return a <= b ? -1 : 1;
+    }
+
+    function convert(stringNumber) {
+      return Number(stringNumber.toLocaleString()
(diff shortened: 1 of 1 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,31 @@
...
 
-// write code here
+const headElements = document
+  .querySelector('thead')
+  .querySelectorAll('th');
+
+for (const item of headElements) {
+  item.addEventListener('click', () => {
+    const table
+      = [...document.querySelector('tbody').querySelectorAll('tr')]
```
Volodymyr-Melnyk-Fron10der, line 10: I suggest to move it into variable
`const bodyRows = document.querySelector('tbody').querySelectorAll('tr')`

src/scripts/main.js
```diff
@@ -1,3 +1,31 @@
...
+        a = (convert(a));
+        b = (convert(b));
+      }
+
+      return a < b ? -1 : 1;
+    }
+
+    function convert(stringNumber) {
+      return Number(stringNumber.toLocaleString().replace(/\D/g, ''));
+    }
```
Volodymyr-Melnyk-Fron10der, line 27: move this function on global level

src/scripts/main.js
```diff
@@ -1,3 +1,31 @@
...
+      let a = aInput.children[item.cellIndex].textContent;
+      let b = bInput.children[item.cellIndex].textContent;
+
+      if (convert(a) > 0) {
+        a = (convert(a));
+        b = (convert(b));
+      }
+
+      return a < b ? -1 : 1;
+    }
```
Volodymyr-Melnyk-Fron10der, line 23: It looks like this function it not belong to here as well. move it on global level

Volodymyr-Melnyk-Fron10der (changes requested):

src/scripts/main.js
Chupiak: I use 'item.cellIndex' inside loop and it must be inside the loop.

Viktor-Kost (approved):
Nice!

PR #558, opened 2023-02-23
Solution

dissolveyo:
- [DEMO LINK](https://dissolveyo.github.io/js_sort_table_DOM/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,92 @@
 'use strict';
 
 // write code here
+const tableHeadRef = document.querySelector('thead');
+const tableBodyRef = document.querySelector('tbody');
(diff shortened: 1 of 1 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,92 @@
...
+        <td>${element.salary}</td>
+      </tr>`
+    )
+    .join('')}
+      `;
+
+  return sortedList;
+}
+
+const compareAsStrings = (a, b) => {
```
loralevitska, line 75: You can use [localeCompare](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/localeCompare) instead custom function

loralevitska (approved):

PR #561, opened 2023-02-24
Solution

VadymTsyndra:
https://VadymTsyndra.github.io/js_sort_table_DOM/

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,65 @@
 'use strict';
 
-// write code here
+const tBody = document.querySelector('tbody');
+
+const data = [...tBody.children];
+
+const titles = document.querySelectorAll('th');
+
+for (const title of Array.from(titles)) {
+  if (title.innerHTML === 'Name') {
+    title.addEventListener('click', (e) => {
+      data.sort((a, b) => {
+        const contentA = a.cells[0].innerHTML;
+        const contentB = b.cells[0].innerHTML;
+
+        return contentA.localeCompare(contentB);
+      });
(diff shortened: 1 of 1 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,65 @@
...
+
+for (const title of Array.from(titles)) {
+  if (title.innerHTML === 'Name') {
+    title.addEventListener('click', (e) => {
+      data.sort((a, b) => {
+        const contentA = a.cells[0].innerHTML;
+        const contentB = b.cells[0].innerHTML;
+
+        return contentA.localeCompare(contentB);
+      });
```
ViktorHavryliuk, line 17: Move sort callback into separate function to reduce code duplication

ViktorHavryliuk (changes requested):
Please use switch case instead of multiple ifs

src/scripts/main.js
```diff
@@ -1,3 +1,34 @@
...
+
+    if (cellIndex === 2) {
+      return contentA - contentB;
+    }
+
+    if (cellIndex === 3) {
+      return getNumber(contentA) - getNumber(contentB);
+    }
+
+    return contentA.localeCompare(contentB);
```
Volodymyr-Melnyk-Fron10der, line 22: use switch/case/default

Volodymyr-Melnyk-Fron10der (changes requested):
last improvement here

Volodymyr-Melnyk-Fron10der (approved):
Well done✅

PR #563, opened 2023-02-26
Solution

EvgenyNikitinEvgeny:
https://EvgenyNikitinEvgeny.github.io/js_sort_table_DOM/

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,61 @@
 'use strict';
 
-// write code here
+const tBody = document.querySelector('tbody');
+const arraySpecialists = [...tBody.children];
+const elementsTotal = document.querySelectorAll('th');
+
+for (const element of elementsTotal) {
+  switch (element.innerText) {
+    case 'Name':
+      element.addEventListener('click', (e) => {
+        arraySpecialists.sort((a, b) => {
+          const itemA = a.cells[0].innerText;
+          const itemB = b.cells[0].innerText;
+
+          return itemA.localeCompare(itemB);
+        });
+
+        return tBody.append(...arraySpecialists);
+      });
+      break;
+    case 'Position':
+      element.addEventListener('click', (e) => {
+        arraySpecialists.sort((a, b) => {
+          const itemA = a.cells[1].innerText;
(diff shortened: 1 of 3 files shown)


EvgenyNikitinEvgeny:
https://evgenynikitinevgeny.github.io/js_sort_table_DOM/

src/scripts/main.js
```diff
@@ -1,3 +1,34 @@
...
-// write code here
+const tBody = document.querySelector('tbody');
+const arraySpecialists = [...tBody.children];
+const elementsTotal = document.querySelectorAll('th');
+
+for (let i = 0; i < elementsTotal.length; i++) {
+  elementsTotal[i].id = [i];
+}
+
+const SortTable = (e) => {
```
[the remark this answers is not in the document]
DarkMistyRoom: use lower cased first letter for simple functions

src/scripts/main.js
```diff
@@ -1,3 +1,34 @@
...
+  arraySpecialists.sort((a, b) => {
+    const itemA = a.cells[e.target.id].innerText;
+    const itemB = b.cells[e.target.id].innerText;
+
+    switch (e.target.innerText) {
+      case 'Name':
+        return itemA.localeCompare(itemB);
+      case 'Position':
+        return itemA.localeCompare(itemB);
```
DarkMistyRoom, line 20: combine these cases

src/scripts/main.js
```diff
@@ -1,3 +1,34 @@
...
+      case 'Age':
+        return itemA - itemB;
+      case 'Salary':
+        return itemA.slice(1).split(',').join('')
+          - itemB.slice(1).split(',').join('');
```
DarkMistyRoom, line 25: use `default `case to handle the invalid sort type. And then let's think: in this default case we still iterate through the whole array. But should we? Let's optimize it. May be would be better to move `switch case` outside the `sort `

DarkMistyRoom, line 25: move this item transformation logic to a separate function to avoid duplicating

DarkMistyRoom (changes requested):

IvanFesenko (approved):
Well done

PR #574, opened 2023-03-06
Ready

sergik94:
https://sergik94.github.io/js_sort_table_DOM/


src/scripts/main.js
```diff
@@ -1,3 +1,34 @@
...
+
+  function getData(employee) {
+    const data = employee.cells[column].innerHTML;
+
+    if (data[0] === '$') {
+      return +data.split(',').join('').slice(1);
+    }
+
+    return data;
+  }
```
IvanFesenko, line 18: No need to create function inside another one
```suggestion

function getData(employee) {
    const data = employee.cells[column].innerHTML;

    if (data[0] === '$') {
      return +data.split(',').join('').slice(1);
    }

    return data;
  }

function sortList(e) {
  const column = e.target.cellIndex;

```

IvanFesenko (changes requested):

src/scripts/main.js
sergik94: @IvanFesenko yes, but in this case, we will have function getData with two parameters (getData(employee, column))

etojeDenys (approved):
LGTM

PR #575, opened 2023-03-07
add task solution

nastychenko:


src/scripts/main.js
```diff
@@ -1,3 +1,43 @@
 'use strict';
 
-// write code here
+const tableData = document.querySelector('table');
+const tableBody = document.querySelector('tbody');
+
+const sortPeople = (targetName) => {
+  if (targetName === 'name') {
+    const sortedRows = [...tableBody.rows]
+      .sort((a, b) => a.cells[0].textContent
```
OleziO, line 9: Make function and use it wherever it's possible: 

`const compareCellText (a, b, i) => a.cells[i].textContent.localeCompare(b.cells[i].textContent)`

OleziO (changes requested):

nazarmatsevych (approved):

PR #579, opened 2023-03-09
Task comleted

read-my-code:
[DEMO LINK](https://read-my-code.github.io/js_sort_table_DOM/)


DarkMistyRoom (approved):
It is a good solution! But I want you to know about another one, which get access to the appropriate `td `of `person `directly without converting it into object.
image

PR #587, opened 2023-03-15
Table sorting by clicking on the title

ilona-romaniuk:
https://ilona-romaniuk.github.io/js_sort_table_DOM/

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -17,7 +17,7 @@ headlines.addEventListener('click', (e) => {
       const checkedRow = convertToNumber(row);
       const checkedNextRow = convertToNumber(nextRow);
 
-      return +checkedRow - +checkedNextRow;
+      return Number(checkedRow) - Number(checkedNextRow);
     }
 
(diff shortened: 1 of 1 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,31 @@
...
+
+  [...listRows].sort((a, b) => {
+    const row = a.cells[index].textContent;
+    const nextRow = b.cells[index].textContent;
+
+    if (/\d/.test(row)) {
+      const checkedRow = convertToNumber(row);
+      const checkedNextRow = convertToNumber(nextRow);
+
+      return +checkedRow - +checkedNextRow;
```
loralevitska, line 20: Try to avoid such expressions `- +`. Use `Number()` in this case https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number
```suggestion
      return Number(checkedRow) - Number(checkedNextRow);
```

loralevitska (approved):

PR #600, opened 2023-03-26
solution

Bohdan33:
- [DEMO LINK](https://bohdan33.github.io/js_sort_table_DOM/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,25 @@
 'use strict';
 
-// write code here
+const thead = document.querySelector('thead tr');
+const table = document.querySelector('tbody');
+const tr = [...table.children];
+
+thead.addEventListener('click', e => {
+  const sorted = tr.sort((a, b) => {
+    const first = a.children[e.target.cellIndex].innerText;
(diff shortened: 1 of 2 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,25 @@
...
+
+thead.addEventListener('click', e => {
+  const sorted = tr.sort((a, b) => {
+    const first = a.children[e.target.cellIndex].innerText;
+    const second = b.children[e.target.cellIndex].innerText;
+
+    switch (e.target.cellIndex) {
+      case 0:
+      case 1:
+        return first.localeCompare(second);
```
OlEzhA0, line 15: this is ok for this task, but this solution is bad for scaling. A good programmer must think ahead. So if you have new columns or the order in which the output columns change, you will need to change the logic code.
I suggest checking not by index, but by content


exmaple

```
if (first[0] === '$') {
  // handle your value and calc it
}

if (isFinite(first)) {
  // it's number value
}

// here you can return `localeCompare`
```

OlEzhA0 (changes requested):

PR #613, opened 2023-04-04
Table sort by salary

serge-kontakt:
- [DEMO LINK](https://serge-kontakt.github.io/js_sort_table_DOM/)

Table is sorting by clicking on headers. But for some reason 3 from 4 tests are failed.

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,49 @@
 'use strict';
 
 // write code here
+
+function toNumber(Str) {
+  const numb = +Str.replace(/\D/g, '');
+
+  return numb;
+}
+
+const sortSalary = () => {
+  const listElements = document.querySelector('tbody').children;
+
+  const arraySort = [...listElements].sort((a, b) => {
+    const salaryA = toNumber(a.lastElementChild.textContent);
+    const salaryB = toNumber(b.lastElementChild.textContent);
+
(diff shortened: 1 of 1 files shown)


Anastasiia-Svintsova (changes requested):
Please, ask questions in the chat if you have them

src/scripts/main.js
```diff
@@ -1,3 +1,62 @@
 'use strict';
 
 // write code here
+
+function toNumber(Str) {
```
polosanya, line 5:
```suggestion
function convertToNumber(str) {
```
Consider naming functions according to these rules
image
There is also no need to capitalize the parameter

src/scripts/main.js
```diff
@@ -1,3 +1,62 @@
...
+  const numb = +Str.replace(/\D/g, '');
+
+  return numb;
+}
+
+const sortAll = (e) => {
+  const arraySort = [...listElements].sort((a, b) => {
+    if (e.target === document.querySelectorAll('th')[0]) {
```
polosanya, line 13: It is better to use a switch...case if you have a limited amount of possible values
https://www.w3schools.com/js/js_switch.asp

polosanya (approved):
Good job! I'm approving it, but please pay attention to the comments)

PR #623, opened 2023-04-12
sort table

2pasha:
[DEMO LINK](https://2pasha.github.io/js_sort_table_DOM/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,26 +1,5 @@
 'use strict';
 
-// const tr = document.querySelectorAll('tr');
-// const list = [];
-
-// tr.forEach(row => {
-//   list.push(row.children[0]);
-//   // console.log(row.children[0].innerHTML);
-// });
-
-// console.log(list);
-
-// function sortList(list) {
-//   const sorted = list.sort((a, b) => {
-//     a.innerText.localeCompare(b.innerText);
-//   });
-
-//   console.log(sorted);
-//   return sorted;
-// }
-
-// console.log(sortList(list));
-
 const table = document.querySelector('table');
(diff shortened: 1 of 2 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,31 @@
...
+});
+
+function sortTable(column) {
+  const rows = Array.from(table.tBodies[0].querySelectorAll('tr'));
+
+  rows.sort((a, b) => {
+    const aValue = a.cells[column.cellIndex].textContent;
+    const bValue = b.cells[column.cellIndex].textContent;
+
+    if (!convertToNumber(aValue)) {
```
mykhalenych, line 19: You are calling `convertToNumber()` twice for the same `aValue` in the sort function. Instead, store the result in a variable and reuse it.

mykhalenych (changes requested):

src/scripts/main.js
```diff
@@ -1,3 +1,33 @@
...
+function sortTable(column) {
+  const rows = Array.from(table.tBodies[0].querySelectorAll('tr'));
+
+  rows.sort((a, b) => {
+    const aValue = a.cells[column.cellIndex].textContent;
+    const bValue = b.cells[column.cellIndex].textContent;
+    const aNumber = convertToNumber(aValue);
+    const bNumber = convertToNumber(bValue);
+
+    if (!aNumber) {
```
etojeDenys, line 21: Instead of checking `!aNumber`, it's better to use `isNaN(aNumber)` for checking if the value is not a number.

etojeDenys (changes requested):
almost done

Viktor-Kost (approved):

PR #625, opened 2023-04-12
solution

mzaloha:
https://mzaloha.github.io/js_sort_table_DOM/

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,36 @@
 'use strict';
 
-// write code here
+const table = document.querySelector('table');
+const tBody = document.querySelector('tbody');
+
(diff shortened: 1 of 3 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,36 @@
 'use strict';
 
-// write code here
+const table = document.querySelector('table');
+const tBody = document.querySelector('tbody');
```
etojeDenys, line 4: Consider using `table.querySelector('tbody')` instead of `document.querySelector('tbody')` to ensure you're selecting the right element inside the table 🧐

etojeDenys (approved):
LGTM

PR #626, opened 2023-04-13
1st version

Grechana:
https://Grechana.github.io/js_sort_table_DOM/

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,35 @@
 'use strict';
 
-// write code here
+const tableHeader = document.querySelector('thead');
+
+const tablebody = document.querySelectorAll('tbody >tr');
+const tbody = document.querySelector('tbody');
+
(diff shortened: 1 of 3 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,35 @@
...
 
-// write code here
+const tableHeader = document.querySelector('thead');
+
+const tablebody = document.querySelectorAll('tbody >tr');
+const tbody = document.querySelector('tbody');
+
+const thead = document.querySelector('thead>tr');
+
+tableHeader.addEventListener('click', (events) => {
```
loralevitska, line 10: The plural in this variable name is confusing because it says that the variable contains an array, but there is an object
```suggestion
tableHeader.addEventListener('click', (e) => {
```

event / e / customEvent etc.

loralevitska (approved):

PR #637, opened 2023-04-22
add task solution

meltourme:
https://meltourme.github.io/js_sort_table_DOM/

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,46 @@
 'use strict';
 
-// write code here
+const thead = document.querySelector('thead');
+const headTableRow = thead.querySelector('tr');
+const tbody = document.querySelector('tbody');
+const tr = [...tbody.querySelectorAll('tr')];
+
+headTableRow.addEventListener('click', (e) => {
+  sortList(tr, [...headTableRow.children].indexOf(e.target));
+});
+
+function sortList(list, i) {
+  switch (i) {
+    case 0:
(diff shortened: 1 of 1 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,46 @@
...
+  let result = '';
+
+  for (const i of item) {
+    if (('0123456789').includes(i)) {
+      result += i;
+    }
+  }
+
+  return +result;
+}
```
ivbutsykin, line 46: Another option to do this:
```suggestion
function toNum(item) {
  return +string.slice(1).split(',').join('');
}
```

src/scripts/main.js
```diff
@@ -1,3 +1,46 @@
 'use strict';
 
-// write code here
+const thead = document.querySelector('thead');
+const headTableRow = thead.querySelector('tr');
+const tbody = document.querySelector('tbody');
+const tr = [...tbody.querySelectorAll('tr')];
+
+headTableRow.addEventListener('click', (e) => {
+  sortList(tr, [...headTableRow.children].indexOf(e.target));
```
ivbutsykin, line 9: Hm, mb better to do this with `e.target.closet('th')`. So when you will receive `Name, Position, ...`. Your variant also can be but it's has a minus because `switch/case` has `0, 1` (not names) as cases.

ivbutsykin (approved):
Good job!

PR #650, opened 2023-05-05
solution

mariuszgit:


src/scripts/main.js
```diff
@@ -1,3 +1,27 @@
...
+  [...allEmployees].sort((prev, next) => {
+    if (prev.children[index].textContent.charAt(0) === '$') {
+      return convertToNumber(prev.children[index].textContent)
+        - convertToNumber(next.children[index].textContent);
+    } else {
+      return (prev.children[index].textContent)
+        .localeCompare(next.children[index].textContent);
+    }
+  }).forEach(item => ulNode.append(item));
+};
```
DorotaLeniecDev, line 17: For readibility I would suggest creating const, also when chaining methods you can start each method from a newline.
Also else statement is not needed
```suggestion
const sortList = (ulNode, index) => {
  [...allEmployees]
    .sort((prev, next) => {
      const prevContent = prev.children[index].textContent;
      const nextContent = next.children[index].textContent;

      if (prevContent.charAt(0) === '$') {
        return convertToNumber(prevContent)- convertToNumber(nextContent);
      } else {
        return (prevContent).localeCompare(nextContent);
      }
    })
  .forEach(item => tableBody.append(item));
```

DorotaLeniecDev (changes requested):
Good job, I left few suggestions to improve your code

DorotaLeniecDev (approved):
Well done :)

PR #653, opened 2023-05-06
sortedList

edkido:
[DEMO LINK](https://edkido.github.io/js_sort_table_DOM/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,32 @@
 'use strict';
 
-// write code here
+const headers = [...document.querySelector('thead').children[0].children];
+const tbody = document.querySelector('tbody');
+const peopleRows = [...document.querySelector('tbody').children];
+
+headers.forEach(head => {
+  const index = headers.indexOf(head);
+
+  head.addEventListener('click', () => {
+    peopleRows.sort((person1, person2) => {
+      let per1 = person1.children[index].innerHTML;
+      let per2 = person2.children[index].innerHTML;
+
+      if (index === 3) {
+        per1 = per1.slice(1).replaceAll(',', '');
+        per2 = per2.slice(1).replaceAll(',', '');
+
+        return per1 - per2;
+      }
+
+      return per1.localeCompare(per2);
+    });
+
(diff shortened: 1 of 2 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,32 @@
...
+        per1 = per1.slice(1).replaceAll(',', '');
+        per2 = per2.slice(1).replaceAll(',', '');
+
+        return per1 - per2;
+      }
+
+      return per1.localeCompare(per2);
+    });
+
+    const copyRows = [...peopleRows];
```
etojeDenys, line 25: The 'copyRows' variable is not necessary, as you are sorting the 'peopleRows' array. You can directly append the sorted items to the 'tbody'.

src/scripts/main.js
```diff
@@ -1,3 +1,32 @@
 'use strict';
 
-// write code here
+const headers = [...document.querySelector('thead').children[0].children];
+const tbody = document.querySelector('tbody');
+const peopleRows = [...document.querySelector('tbody').children];
+
+headers.forEach(head => {
+  const index = headers.indexOf(head);
```
etojeDenys, line 8: The 'indexOf' method is not efficient here, use the second parameter of 'forEach' to get the index.

src/scripts/main.js
```diff
@@ -1,3 +1,32 @@
 'use strict';
 
-// write code here
+const headers = [...document.querySelector('thead').children[0].children];
```
etojeDenys, line 3: Instead of using 'children[0].children', use 'querySelectorAll' to get all the 'th' elements.

src/scripts/main.js
```diff
@@ -1,3 +1,32 @@
...
+const tbody = document.querySelector('tbody');
+const peopleRows = [...document.querySelector('tbody').children];
```
etojeDenys, line 5: Instead of using 'children', use 'querySelectorAll' to get all the 'tr' elements from 'tbody'.

etojeDenys (changes requested):

VitaliyBondarenko1982 (approved):

PR #656, opened 2023-05-09
add task MOROZ v1

morozit:
https://morozit.github.io/js_sort_table_DOM/

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -4,30 +4,24 @@ const table = document.querySelector('table');
 const thead = table.querySelector('thead tr');
 const tbody = table.querySelector('tbody');
 const rows = tbody.querySelectorAll('tr');
-const number = string => string.replace(/[$,]/g, '');
+const formatSalary = string => string.replace(/[$,]/g, '');
 
 thead.addEventListener('click', even => {
   const headerIndex = Array.from(thead.children).indexOf(even.target);
+  const sortRows = Array.from(rows);
 
-  if (even.target.innerHTML === 'Salary') {
-    Array.from(rows)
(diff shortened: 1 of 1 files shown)


anstsot (changes requested):
It seems the sorting by salary isn't working correct. Could you fix it please?
image

You can use `switch/case` to add different sortings depending on type (numeric or alphabetic)

src/scripts/main.js
```diff
@@ -1,3 +1,33 @@
 'use strict';
 
-// write code here
+const table = document.querySelector('table');
+const thead = table.querySelector('thead tr');
+const tbody = table.querySelector('tbody');
+const rows = tbody.querySelectorAll('tr');
+const number = string => string.replace(/[$,]/g, '');
```
polosanya, line 7: Consider renaming the 'number' function to a more descriptive name, e.g. 'formatSalary'.

src/scripts/main.js
```diff
@@ -1,3 +1,33 @@
...
+      });
+  } else {
+    Array.from(rows)
+      .sort((a, b) => {
+        const aValue = a.children[headerIndex].textContent;
+        const bValue = b.children[headerIndex].textContent;
+
+        return aValue.localeCompare(bValue);
+      }).forEach(row => {
+        tbody.appendChild(row);
```
polosanya, line 30: Actually, there is no need to have two totally different logics for salary and other values, as many moments are the same for all columns. Just provide only different sorting if it's the salary column

polosanya (changes requested):
Good job! Let's improve your code a bit)

agniya-i (approved):

PR #661, opened 2023-05-15
add task solution

NikaNika12:
[DEMO LINK](https://NikaNika12.github.io/js_sort_table_DOM/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,30 @@
 'use strict';
 
-// write code here
+const headTable = document.querySelector('thead');
+const bodyTable = document.querySelector('tbody');
+const rows = [...bodyTable.rows];
+
+headTable.addEventListener('click', e => {
+  if (!e.target.tagName === 'th') {
+    return;
+  };
+
+  const targetIndex = e.target.cellIndex;
+
+  function numberOptimizer(str) {
+    return str.replace(/[$,]/g, '');
+  };
+
+  const sortList = rows.sort((a, b) => {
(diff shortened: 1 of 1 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,30 @@
...
+
+  function numberOptimizer(str) {
+    return str.replace(/[$,]/g, '');
+  };
+
+  const sortList = rows.sort((a, b) => {
+    const rowOne = numberOptimizer(a.cells[targetIndex].textContent);
+    const rowTwo = numberOptimizer(b.cells[targetIndex].textContent);
+
+    if (typeof elementOne === 'number') {
```
mykhalenych, line 22: what is elementOne?
I don't see this variable

mykhalenych (changes requested):

src/scripts/main.js
```diff
@@ -1,3 +1,30 @@
...
+  function numberOptimizer(str) {
+    return str.replace(/[$,]/g, '');
+  };
+
+  const sortList = rows.sort((a, b) => {
+    const rowOne = numberOptimizer(a.cells[targetIndex].textContent);
+    const rowTwo = numberOptimizer(b.cells[targetIndex].textContent);
+
+    if (typeof rowOne === 'number') {
+      return rowOne - rowTwo;
```
TarasHoliuk, line 23: It still compares as `string`, because everything in `textContent` always is `string`. You should try to transform it into `number` and check if it's valid (`isNaN`)

image

TarasHoliuk (changes requested):

tetiana-ivanchenko (approved):

PR #670, opened 2023-05-19
Add solution

ruslanpashkov:
[DEMO LINK](https://ruslanpashkov.github.io/js_sort_table_DOM/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,44 @@
 'use strict';
 
-// write code here
+const table = document.querySelector('table');
+const tableRows = table.querySelectorAll('tr');
+const people = [...tableRows].slice(1, -1);
+
+table.firstElementChild.addEventListener('click', (e) => {
+  people.sort((personA, personB) => {
+    switch (e.target.innerText) {
+      case 'Name':
+        const nameA = personA.children[0].innerText;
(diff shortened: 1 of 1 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,44 @@
...
+        const salaryB = personB.children[3].innerText;
+
+        return extractDigits(salaryA) - extractDigits(salaryB);
+
+      default:
+        throw new Error(`Unknown table header: ${e.target.innerText}`);
+    }
+  });
+
+  table.append(...people);
```
33Orange, line 39: Appending sorted people should be inside the event listener callback, after the sort function.

src/scripts/main.js
```diff
@@ -1,3 +1,44 @@
 'use strict';
 
-// write code here
+const table = document.querySelector('table');
+const tableRows = table.querySelectorAll('tr');
+const people = [...tableRows].slice(1, -1);
+
+table.firstElementChild.addEventListener('click', (e) => {
```
33Orange, line 7: It's better to use event delegation on 'th' elements, rather than on the whole 'thead'. Consider changing the event listener target to 'th' elements.

33Orange (approved):
Well done !

PR #676, opened 2023-05-24
solution

IzzetYusufov:
[DEMO LINK](https://IzzetYusufov.github.io/js_sort_table_DOM/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,32 @@
 'use strict';
 
-// write code here
+const headers = document.querySelector('thead');
+const tbody = document.querySelector('tbody');
+const rows = tbody.rows;
+
+headers.addEventListener('click', e => {
+  const target = e.target;
+  const headerList = headers.firstElementChild.children;
+  const index = [...headerList].findIndex(tag => tag === target);
+  let sort;
+
+  if (e.target.innerText === 'Salary') {
+    sort = [...rows].sort((a, b) => {
+      const start = a.children[index].innerText.slice(1).replace(',', '.');
+      const end = b.children[index].innerText.slice(1).replace(',', '.');
+
+      return end - start;
+    });
+  } else {
+    sort = [...rows].sort((a, b) => {
+      const start = a.children[index].innerText;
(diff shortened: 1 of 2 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,32 @@
...
+      const start = a.children[index].innerText;
+      const end = b.children[index].innerText;
+
+      return end.localeCompare(start);
+    });
+  }
+
+  for (const element of sort) {
+    tbody.prepend(element);
+  }
```
Viktor-Kost, line 31: apply forEach method instead of for of loop
`sort.forEach`...

src/scripts/main.js
```diff
@@ -1,3 +1,32 @@
...
-// write code here
+const headers = document.querySelector('thead');
+const tbody = document.querySelector('tbody');
+const rows = tbody.rows;
+
+headers.addEventListener('click', e => {
+  const target = e.target;
+  const headerList = headers.firstElementChild.children;
+  const index = [...headerList].findIndex(tag => tag === target);
+  let sort;
```
Viktor-Kost, line 11:
```suggestion
```

src/scripts/main.js
```diff
@@ -1,3 +1,32 @@
...
+      return end - start;
+    });
+  } else {
+    sort = [...rows].sort((a, b) => {
+      const start = a.children[index].innerText;
+      const end = b.children[index].innerText;
+
+      return end.localeCompare(start);
+    });
+  }
```
Viktor-Kost, line 27: try like this 

```
  const sort = [...rows].sort((a, b) => {
    const start = a.children[index].innerText;
    const end = b.children[index].innerText;

    if (target.innerText === 'Salary') {
      const parseSalary = salary => parseFloat(salary.slice(1).replace(',', '.'));
      return parseSalary(end) - parseSalary(start);
    } else {
      return end.localeCompare(start);
    }
  });
```

Viktor-Kost (changes requested):

vlad-demchuk (approved):

PR #682, opened 2023-05-27
Solution

AV1788:
[DEMO LINK](https://AV1788.github.io/js_sort_table_DOM/)


src/scripts/main.js
```diff
@@ -1,3 +1,58 @@
...
+const tfootThs = tfoot.querySelectorAll('th');
+
+function convNumber(text) {
+  return Number(text.replace(/[^0-9.-]+/g, ''));
+}
+
+table.addEventListener('click', (e) => {
+  const item = e.target;
+
+  for (const th of tfootThs) {
```
TarasHoliuk, line 16: add the event listener to `thead` and get rid of this check

src/scripts/main.js
```diff
@@ -1,3 +1,58 @@
...
+    if (item === th) {
+      return;
+    }
+  }
+
+  if (item === theadThs[0]) {
```
TarasHoliuk, line 22: use `swith` block (check `cellIndex` of target) instead of multiple `if` blocks

src/scripts/main.js
```diff
@@ -1,3 +1,58 @@
...
+    const sortedRows = Array.from(table.rows)
+      .slice(1, -1)
+      .sort((rowA, rowB) =>
+        rowA.cells[0].innerHTML > rowB.cells[0].innerHTML ? 1 : -1);
+
+    table.tBodies[0].append(...sortedRows);
```
TarasHoliuk, line 28: you are doing almost the same in each case

add `switch` block inside `sort` callback to avoid duplicating code
Hint: you may join some cases if you handle them in the same way:
```
case 1:
case 2:
  // you way to handle these cases
  ```

File: src/scripts/main.js [hunk 4]
```diff
@@ -1,3 +1,58 @@
...
+  }
+
+  if (item === theadThs[3]) {
+    const sortedRows3 = Array.from(table.rows)
+      .slice(1, -1)
+      .sort((rowA, rowB) =>
+        convNumber(rowA.cells[3].innerHTML)
+        - convNumber(rowB.cells[3].innerHTML));
+
+    table.tBodies[0].append(...sortedRows3);
```
TarasHoliuk, line 56: after changes I've described above you should do it once

TarasHoliuk (changes requested):

TarasHoliuk (changes requested):

src/scripts/main.js
```diff
@@ -1,3 +1,29 @@
...
+function convNumber(text) {
+  return Number(text.replace(/[^0-9.-]+/g, ''));
+}
+
+thead.addEventListener('click', (elem) => {
+  const item = elem.target;
+  const sortedRows = Array.from(table.rows)
+    .slice(1, -1).sort((a, b) => {
+      switch (item.cellIndex) {
+        case 0:
```
OleziO, line 15: Try to move all cases to the object.

src/scripts/main.js
```diff
@@ -1,3 +1,29 @@
...
+        case 1:
+          return a.cells[item.cellIndex].innerHTML
+          > b.cells[item.cellIndex].innerHTML ? 1 : -1;
+        case 2:
+          return Number(a.cells[item.cellIndex].innerHTML)
```
OleziO, line 20: If you have your own function that converts something to num - better is to use it.

OleziO (changes requested):

TarasHoliuk (approved):
Well done :+1:

PR #686, opened 2023-06-01
add task solution

antonina-klishch:
https://antonina-klishch.github.io/js_sort_table_DOM/

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,26 @@
 'use strict';
 
 // write code here
+
+const body = document.querySelector('tbody');
+const rows = body.querySelectorAll('tr');
+const title = document.querySelector('thead');
+
+title.addEventListener('click', (e) => {
+  const index = e.target.cellIndex;
+  const sorted = Array.from(rows).sort((A, B) => {
+    const cellA = A.children[index].textContent;
+    const cellB = B.children[index].textContent;
+
+    if (index === 3) {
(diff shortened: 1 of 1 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,26 @@
...
 
 // write code here
+
+const body = document.querySelector('tbody');
+const rows = body.querySelectorAll('tr');
+const title = document.querySelector('thead');
+
+title.addEventListener('click', (e) => {
+  const index = e.target.cellIndex;
+  const sorted = Array.from(rows).sort((A, B) => {
```
VitaliyBodnar, line 11: don't use the variable or parameters in LowerCase, you need to use CamelCame for that

src/scripts/main.js
```diff
@@ -1,3 +1,26 @@
...
+    const cellA = A.children[index].textContent;
+    const cellB = B.children[index].textContent;
+
+    if (index === 3) {
```
VitaliyBodnar, line 15: try to don't use the 'magic number' in the code. If you want to use index 3, you need to decorate it to the cost.

src/scripts/main.js
VitaliyBodnar, line 11: better to use [...rows] instead of Array.from(rows), but it's not required

VitaliyBodnar (changes requested):
Well done 🔥
Let's improve your solution 🧑‍💻

sTorba24 (approved):
GJ 🚀

PR #701, opened 2023-07-20
add task solution

Luk2asz:
- [DEMO LINK](https://Luk2asz.github.io/js_sort_table_DOM/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,52 @@
 'use strict';
 
-// write code here
+const mainRow = document.querySelector('tr');
+const cells = mainRow.querySelectorAll('th');
+const tableBody = document.querySelector('tbody');
+const tableRow = tableBody.querySelectorAll('tr');
+const rowArray = Array.from(tableRow);
+
+cells.forEach(cell => {
+  cell.addEventListener('click', function() {
+    if (cell.textContent === 'Name') {
+      rowArray.sort((a, b) => {
+        const firstElement = a.children[0].textContent;
+        const secondElement = b.children[0].textContent;
+
+        return firstElement.localeCompare(secondElement);
+      });
+
+      rowArray.forEach(x => tableBody.appendChild(x));
+    } else if (cell.textContent === 'Position') {
+      rowArray.sort((a, b) => {
+        const firstElement = a.children[1].textContent;
+        const secondElement = b.children[1].textContent;
+
+        return firstElement.localeCompare(secondElement);
+      });
+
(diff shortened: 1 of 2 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,52 @@
...
 
-// write code here
+const mainRow = document.querySelector('tr');
+const cells = mainRow.querySelectorAll('th');
+const tableBody = document.querySelector('tbody');
+const tableRow = tableBody.querySelectorAll('tr');
+const rowArray = Array.from(tableRow);
+
+cells.forEach(cell => {
+  cell.addEventListener('click', function() {
```
DorotaLeniecDev, line 10: Add event listener to thead (check [event delegation](https://www.freecodecamp.org/news/event-delegation-javascript/)), then you can use `cellIndex` and in your sort function you can one time create `current` and `next` consts and later use those values to sort the table, based on its type. No need to repeate this for each case. 

```js
const thead = document.querySelector('thead');
const tbody = table.querySelector('tbody');
const rows = [...tbody.querySelectorAll('tr')];

thead.addEventListener('click', e => {
  const columnIndex = e.target.cellIndex;

  rows.sort((a, b) => {
    const current = a.cells[cellIndex].innerHTML;
    const next = b.cells[cellIndex].innerHTML;

   // sort based on the type, 
   // for Age and Salary 
   // and fot the rest you can use localCompare

}
```

DorotaLeniecDev (changes requested):
Need some improvements

src/scripts/main.js
```diff
@@ -1,3 +1,27 @@
...
+    const current = a.cells[columnIndex].innerHTML;
+    const next = b.cells[columnIndex].innerHTML;
+
+    if (e.target.cellIndex === 0 || e.target.cellIndex === 1) {
+      return current.localeCompare(next);
+    } else if (e.target.cellIndex === 2) {
+      return Number(current) - Number(next);
+    } else if (e.target.cellIndex === 3) {
+      return salaryToNumber(current) - salaryToNumber(next);
+    }
```
DorotaLeniecDev, line 20: This could be writen witoud else if
```suggestion

if (columnIndex === 2) {
  return Number(current) - Number(next);
}

if (columnIndex === 3) {
  return salaryToNumber(current) - salaryToNumber(next);
}

return current.localeCompare(next);
```

DorotaLeniecDev (changes requested):
Good job, one last change needed

DorotaLeniecDev (approved):
Good job 👌

PR #702, opened 2023-07-28
sort_table_DOM - first try

RufatRM:
https://RufatRM.github.io/js_sort_table_DOM/

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,37 @@
 'use strict';
 
-// write code here
+const tableBody = document.querySelector('tbody');
+const rowsArray = Array.from(document.querySelectorAll('tbody tr'));
+const headings = document.querySelector('thead');
+const headingsArray = [];
+
+Array.from(headings.firstElementChild.children).forEach(heading => {
+  headingsArray.push(heading.textContent);
+});
+
+function convertToNumber(string) {
+  return parseInt(string.slice(1).replaceAll(',', ''));
+}
+
+headings.addEventListener('click', e => {
+  const index = headingsArray.indexOf(e.target.textContent);
+
+  rowsArray.sort((a, b) => {
+    const aTextContent = a.cells[index].textContent;
+    const bTextContent = b.cells[index].textContent;
+
+    if (+aTextContent) {
+      return +aTextContent - (+bTextContent);
+    } else if (+convertToNumber(aTextContent)) {
+      return convertToNumber(aTextContent)
+      - convertToNumber(bTextContent);
+    } else if (typeof aTextContent === 'string'
+    && !convertToNumber(aTextContent)) {
+      return aTextContent.localeCompare(bTextContent);
+    }
+  });
+
+  rowsArray.forEach(row => {
+    tableBody.appendChild(row);
+  });
+});


src/scripts/main.js
```diff
@@ -1,3 +1,37 @@
...
+  const index = headingsArray.indexOf(e.target.textContent);
+
+  rowsArray.sort((a, b) => {
+    const aTextContent = a.cells[index].textContent;
+    const bTextContent = b.cells[index].textContent;
+
+    if (+aTextContent) {
+      return +aTextContent - (+bTextContent);
+    } else if (+convertToNumber(aTextContent)) {
+      return convertToNumber(aTextContent)
```
lerastarynets, line 26: You are calling the same function convertToNumber twice for 'aTextContent' and 'bTextContent'. To improve performance, you should store the result of these function calls in a variable and use them in your operations.

src/scripts/main.js
```diff
@@ -1,3 +1,37 @@
...
+Array.from(headings.firstElementChild.children).forEach(heading => {
+  headingsArray.push(heading.textContent);
+});
+
+function convertToNumber(string) {
+  return parseInt(string.slice(1).replaceAll(',', ''));
+}
+
+headings.addEventListener('click', e => {
+  const index = headingsArray.indexOf(e.target.textContent);
```
lerastarynets, line 17: Instead of finding the index of the clicked header in an array, it would be more efficient to use the cellIndex property of the clicked cell. This would simplify your code and improve performance.

src/scripts/main.js
```diff
@@ -1,3 +1,37 @@
...
+      - convertToNumber(bTextContent);
+    } else if (typeof aTextContent === 'string'
+    && !convertToNumber(aTextContent)) {
+      return aTextContent.localeCompare(bTextContent);
+    }
+  });
+
+  rowsArray.forEach(row => {
+    tableBody.appendChild(row);
+  });
```
lerastarynets, line 36:
```suggestion
  tableBody.append(...rowsArray);
  });
```

lerastarynets (changes requested):
Nice start! Let's fix and improve your work a little bit)

src/scripts/main.js
```diff
@@ -1,3 +1,32 @@
...
+
+  rowsArray.sort((a, b) => {
+    const aText = a.cells[index].textContent;
+    const bText = b.cells[index].textContent;
+    const callWithA = convertToNumber(aText);
+    const callWithB = convertToNumber(bText);
+
+    if (+aText) {
+      return +aText - (+bText);
+    } else if (+callWithA) {
```
Anastasiia-Svintsova, line 22: You've converted it to a number already 
```suggestion
    } else if (callWithA) {
```

src/scripts/main.js
```diff
@@ -1,3 +1,32 @@
...
+      return callWithA
+      - callWithB;
+    } else if (typeof aText === 'string'
+    && !callWithA) {
```
Anastasiia-Svintsova, line 26:
```suggestion
    } else if (typeof aText === 'string' && !callWithA) {
```

Anastasiia-Svintsova (approved):

PR #705, opened 2023-08-03
sol

GGsmou:
[DEMO LINK](https://GGsmou.github.io/js_sort_table_DOM/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,5 +1,8 @@
 'use strict';
 
+const getSalary = salaryElement =>
+  parseInt(salaryElement.innerText.slice(1).replace(',', ''));
+
 const thead = document.querySelector('thead').children[0];
 
 thead.addEventListener('click', e => {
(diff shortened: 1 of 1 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,45 @@
...
+          b.children[2].innerText
+        );
+      case 'Salary':
+        const aSalary = parseInt(
+          a.children[3].innerText.slice(1).replace(',', '')
+        );
+
+        const bSalary = parseInt(
+          b.children[3].innerText.slice(1).replace(',', '')
+        );
```
OlEzhA0, line 32: create function `getSalary` that will accept only `a.children[3]` or `b.children[3]` and return parsed, sliced and replaced value

src/scripts/main.js
```diff
@@ -1,3 +1,45 @@
...
+
+  const sortType = target.innerText;
+  const data = Array.from(document.querySelectorAll('tbody tr'));
+
+  data.sort((a, b) => {
+    switch (sortType) {
+      case 'Name':
+        return a.children[0].innerText.localeCompare(b.children[0].innerText);
+      case 'Position':
+        return a.children[1].innerText.localeCompare(b.children[1].innerText);
```
OlEzhA0, line 20: as I see, you can combine it in 1 case
```
      const caseConfig = {
        name: 'Name',
        position: 'Position',
      }

      ...
      case caseConfig.name:
      case caseConfig.position:
        const idxConfig = { [caseConfig.name]: 0, [caseConfig.position]: 1 }
        const idx = idxConfig[sortType]

        return a.children[idx].innerText.localeCompare(b.children[idx].innerText);
```

OlEzhA0 (changes requested):

src/scripts/main.js
```diff
@@ -1,3 +1,55 @@
...
+  };
+
+  data.sort((a, b) => {
+    switch (sortType) {
+      case CASES.name:
+      case CASES.position:
+        const idxCases = {
+          [CASES.name]: 0,
+          [CASES.position]: 1,
+        };
```
DanilWeda, line 32:
```suggestion
        const idx = sortType === CASES.name ? 0 : 1
```

DanilWeda (approved):
Fine work!

PR #708, opened 2023-08-22
Solution

ArtemVelychko:
- [DEMO LINK](https://ArtemVelychko.github.io/js_sort_table_DOM/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,30 @@
 'use strict';
 
-// write code here
+const thead = document.querySelector('thead');
+const mainColumn = thead.querySelector('tr');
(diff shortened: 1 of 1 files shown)


anastasiia-tilikina (changes requested):
sorting doesn't work( it sorts only by salary.
clicking on other headings does nothing. please fix it.

and paste your demo link correctly to the description.

anastasiia-tilikina (changes requested):
Great!

But let's have a look at your listeners. they share a similar logic. let's combine them and use [event delegation](https://javascript.info/event-delegation) pattern.

witflash (approved):
Good job 👍

PR #709, opened 2023-08-25
Develop

Spec0s:
https://spec0s.github.io/js_sort_table_DOM/


anastasiia-tilikina (changes requested):
Great job! 
But we can simplify the sort method. Basically, we have only two methods of sorting: strings and numbers. let's fix it. you can add additional parameters and move appending rows outside the sort method.

anastasiia-tilikina (approved):
well done!

PR #723, opened 2023-10-24
Solution

andrii-rykal:
[DEMO LINK](https://andrii-rykal.github.io/js_sort_table_DOM/)


src/scripts/main.js
```diff
@@ -1,3 +1,53 @@
 'use strict';
 
-// write code here
+document.addEventListener('click', e => {
+  if (e.target.nodeName !== 'TH') {
+    return;
+  };
```
VitaliyBodnar, line 6: it's better to set click only for th elements instead of a whole document

src/scripts/main.js
```diff
@@ -1,3 +1,53 @@
...
+
+  const theadElements = [...document
+    .querySelector('thead').firstElementChild.children];
```
VitaliyBodnar, line 9:
```suggestion
  const theadElements = [
   ...document.querySelector('thead').firstElementChild.children
  ];
```

src/scripts/main.js
```diff
@@ -1,3 +1,53 @@
...
+function sortBy(tbEl, thEl, value) {
+  for (let i = 0; i < thEl.length; i++) {
+    if (thEl[i].textContent !== value) {
+      continue;
+    }
+
+    tbEl.sort((a, b) => {
+      switch (value) {
+        case 'Name':
+        case 'Position':
```
VitaliyBodnar, line 38: it's better to move Name, Position, Age and Salary to the constants as it will make a code cleaner

VitaliyBodnar (changes requested):

VitaliyBodnar (approved):

PR #738, opened 2023-12-05
solution

donizer:
[DEMO LINK](https://donizer.github.io/js_sort_table_DOM/)


VitaliyBodnar (changes requested):
- you need to sort table every time when header name is clicked
- if list is sorted you need to reverse it
https://github.com/mate-academy/js_sort_table_DOM/assets/64746943/00b94b2e-8398-40f3-a36f-2721e6531ea0

lerastarynets (approved):
You're making excellent progress!

PR #756, opened 2024-02-05
solution

Arkadiuszx:
[DEMO LINK](https://Arkadiuszx.github.io/js_sort_table_DOM/)


yevhenii-pyl (approved):
GJ, looks like you went above the task requirements and also implemented a DESC sorting. 

I would probably change the `const bySalary` to => `isSortingBySalary` but that's more of an advice here.

PR #771, opened 2024-03-29
add task solution

valeriaoleynik:
[DEMO LINK](https://valeriaoleynik.github.io/js_sort_table_DOM/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,57 @@
 'use strict';
 
-// write code here
+const table = document.querySelector('table');
+const thList = document.querySelectorAll('th');
+
+thList.forEach((th) => {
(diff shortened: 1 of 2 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,57 @@
...
+  const thName = th.dataset.name;
+
+  sortTable(thIndex, thName);
+});
+
+function sortTable(colNum, type) {
+  const tbody = document.querySelector('tbody');
+  const rowsArray = Array.from(tbody.rows);
+
+  switch (type) {
```
witflash, line 26: As the best practice, the `switch/case` block should have a `default` statement. It may return `null` or some default value.

witflash (approved):
Good job! 👍🏻

PR #773, opened 2024-04-03
Solution

aMtiIV:
[DEMO LINK](https://aMtiIV.github.io/js_sort_table_DOM/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,93 @@
 'use strict';
 
-// write code here
+function addCommasEveryThreeChars(text) {
+  if (text.length <= 3) {
+    return text;
+  }
+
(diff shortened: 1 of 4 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,93 @@
...
+  switch (tableEvent.target.innerText) {
+    case 'Name':
+      comparator = (firstPerson, secondPerson) => {
+        if (firstPerson.name > secondPerson.name) {
+          return 1;
+        } else if (firstPerson.name < secondPerson.name) {
+          return -1;
+        } else {
+          return 0;
+        }
```
BogdanMaliuta, line 48: dont need to use else if, when you are using return

BogdanMaliuta (approved):
Hello, nice job.
But try to avoid reduntant else/if when you using returns, also try remove recursion in your code, just simplify this func

PR #784, opened 2024-05-16
solution

katya-sn:
https://katya-sn.github.io/js_sort_table_DOM/


volodymyr-soltys97 (changes requested):
Hi, where is the styles? You need to make a deploy and check it after on the demo link
image

Task description 
image

lerastarynets (approved):
lgtm 🚀

PR #809, opened 2024-06-18
solution

getchards:
[DEMO LINK](https://getchards.github.io/js_sort_table_DOM/)


DarkMistyRoom (changes requested):
Great job! Only one suggestion: You use one common direction variable for all columns. Would be better to make it always sort ascending on the first click on a column

vadiimvooo (approved):
Well done

PR #822, opened 2024-06-22
sort table by clicking on table titles

mikezhylka:
Hello! I don't know if it is a correct solution, but I think table will look better if sorted elements do not mutate to previous ones after clicking on another title
 
- [DEMO LINK](https://mikezhylka.github.io/js_sort_table_DOM/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,73 @@
 'use strict';
 
-// write code here
+const titles = [...document.querySelectorAll('thead tr th')];
+
+const nameTitle = titles[0];
+const positionTitle = titles[1];
+const ageTitle = titles[2];
+const salaryTitle = titles[3];
+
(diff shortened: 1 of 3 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,73 @@
 'use strict';
 
-// write code here
+const titles = [...document.querySelectorAll('thead tr th')];
+
+const nameTitle = titles[0];
+const positionTitle = titles[1];
+const ageTitle = titles[2];
+const salaryTitle = titles[3];
```
VitaliyBondarenko1982, line 8:
```suggestion
const [nameTitle, positionTitle, ageTitle, salaryTitle] = titles;
```
NI
better to use destructuring

VitaliyBondarenko1982 (approved):
Looks good

PR #836, opened 2024-07-02
solution

VerJane23:
https://verjane23.github.io/js_sort_table_DOM/


maxim2310 (changes requested):
Table must sort rows when i click on headers, plz check it

lerastarynets (approved):
You're on the right track!

PR #841, opened 2024-07-04
add task solution

Svimpik:
https://Svimpik.github.io/js_sort_table_DOM/


etojeDenys (approved):
good job, but it's bad idea to sort table by index, better to sort it by text or smth like this

PR #849, opened 2024-07-08
Solution

Galers:
Have a nice day!😊😊😊
[DEMO LINK](https://Galers.github.io/js_sort_table_DOM/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,37 @@
 'use strict';
 
-// write code here
+const table = document.querySelector('table');
+const headers = table.querySelectorAll('th');
+const rows = Array.from(table.querySelectorAll('tbody tr'));
+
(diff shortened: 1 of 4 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,37 @@
 'use strict';
 
-// write code here
+const table = document.querySelector('table');
+const headers = table.querySelectorAll('th');
+const rows = Array.from(table.querySelectorAll('tbody tr'));
+
+headers.forEach((header, columnIndex) => {
```
TarasHoliuk, line 7: instead of adding an event listener to each header separately it's better to use [event delegation](https://www.freecodecamp.org/news/event-delegation-javascript/). Event bubbling allows to do it

FYI

TarasHoliuk (approved):
GJ 👍 
Added 1 comment FYI 🙂

PR #858, opened 2024-07-12
solution

SerhiiVoitiuk:
- [DEMO LINK](https://SerhiiVoitiuk.github.io/js_sort_table_DOM/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,80 @@
 'use strict';
 
-// write code here
+const table = document.querySelector('table');
+
+table.addEventListener('click', (e) => {
+  const bodies = document.querySelector('tbody');
+  const bodyRows = [...bodies.rows];
+
+  const data = bodyRows.map((el) => {
+    const newObject = {
+      name: el.cells[0].textContent,
+      position: el.cells[1].textContent,
+      age: el.cells[2].textContent,
+      salary: el.cells[3].textContent,
+    };
+
+    return newObject;
+  });
+
+  bodies.remove();
+
+  const newBodies = document.createElement('tbody');
+
+  table.tHead.after(newBodies);
+
+  let sortedData = [];
+
+  switch (e.target.textContent) {
+    case 'Name': {
+      sortedData = [...data].sort((el1, el2) => {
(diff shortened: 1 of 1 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,80 @@
...
+
+  const newBodies = document.createElement('tbody');
+
+  table.tHead.after(newBodies);
+
+  let sortedData = [];
+
+  switch (e.target.textContent) {
+    case 'Name': {
+      sortedData = [...data].sort((el1, el2) => {
```
GUSILLUS, line 30: Consider creating variable with `[...data]` and avoid to make a copy action in every case

src/scripts/main.js
```diff
@@ -1,3 +1,80 @@
...
+          +el1.salary.slice(1).split(',').join('') -
+          +el2.salary.slice(1).split(',').join(''),
+      );
+      break;
+    }
+
+    default: {
+      sortedData = [...data];
+    }
+  }
```
GUSILLUS, line 62: Also in all cases, you use a sort method, so can we place a switch case inside the sort method and change a callback return result?

src/scripts/main.js
```diff
@@ -1,3 +1,80 @@
 'use strict';
 
-// write code here
+const table = document.querySelector('table');
+
+table.addEventListener('click', (e) => {
+  const bodies = document.querySelector('tbody');
+  const bodyRows = [...bodies.rows];
+
+  const data = bodyRows.map((el) => {
```
GUSILLUS, line 9: Consider changing `el` to `row` to improve readability

src/scripts/main.js
```diff
@@ -1,3 +1,80 @@
...
+      );
+      break;
+    }
+
+    default: {
+      sortedData = [...data];
+    }
+  }
+
+  const newElements = sortedData.map((el) => {
```
GUSILLUS, line 64: Same here `el` to `row` 
`newElements` to `newRows` or `updatedRows`

GUSILLUS (changes requested):
GJ!

vadiimvooo (approved):
Well done

PR #895, opened 2024-08-06
v 1.0

maxproway:
- [DEMO LINK](https://maxproway.github.io/js_sort_table_DOM/)


volodymyr-soltys97 (approved):
Good job 👍 
But you don't need to insert the link as a block of code, it should be as plain text
image

PR #906, opened 2024-08-13
Solution

overnall:
- [DEMO LINK](https://<your_account>.github.io/js_sort_table_DOM/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,69 @@
 'use strict';
 
-// write code here
+const td = document.querySelectorAll('td');
+const arr = {};
+let counter = 0;
+
+for (let i = 0; i < td.length / 4; i++) {
+  createRows(i, counter, counter + 4);
+  counter += 4;
+}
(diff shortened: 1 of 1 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,69 @@
...
+document.addEventListener('click', (e) => {
+  if (e.target.textContent === 'Name') {
+    sortTable(0);
+  } else if (e.target.textContent === 'Position') {
+    sortTable(1);
+  } else if (e.target.textContent === 'Age') {
+    sortTable(2);
+  } else if (e.target.textContent === 'Salary') {
+    sortTable(3);
+  }
```
volodymyr-soltys97, line 21: Use switch case here instead of if/ else if conditions

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's make your code better

GUSILLUS (approved):
GJ!

PR #956, opened 2024-09-19
My implemention the task

Serhii-Khobotov:
https://Serhii-Khobotov.github.io/js_sort_table_DOM/

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,36 @@
 'use strict';
 
-// write code here
+const table = document.querySelector('table');
+const headers = table.querySelectorAll('th');
+
+headers.forEach((header, index) => {
+    header.addEventListener('click', () => {
+        setSortTable(index);
+    });
+});
+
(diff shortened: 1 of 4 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,36 @@
...
+
+headers.forEach((header, index) => {
+    header.addEventListener('click', () => {
+        setSortTable(index);
+    });
+});
+
+const setSortTable = (columnIndex) => {
+    const tBody = table.querySelector('tbody');
+    const rows = tBody.querySelectorAll('tr');
```
anastasiiavorobiova: Consider creating an array from the nodelist before using array methods

anastasiiavorobiova (changes requested):
Great work! Almost done!

src/scripts/main.js
Serhii-Khobotov: image

lerastarynets (approved):
Keep up the great work!

PR #985, opened 2024-10-10
Develop

Krykunov:
- [DEMO LINK](https://Krykunov.github.io/js_sort_table_DOM/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -14,7 +14,7 @@ headers.forEach((header, i) => {
       const cellA = a.querySelectorAll('td')[i].textContent;
       const cellB = b.querySelectorAll('td')[i].textContent;
 
-      if (numberFormat(cellA) - numberFormat(cellB)) {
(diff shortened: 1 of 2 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,26 @@
...
+
+const numberFormat = (n) => Number(n.replace(/[^0-9.-]+/g, ''));
+
+headers.forEach((header, i) => {
+  header.addEventListener('click', () => {
+    rows.sort((a, b) => {
+      const cellA = a.querySelectorAll('td')[i].textContent;
+      const cellB = b.querySelectorAll('td')[i].textContent;
+
+      if (numberFormat(cellA) - numberFormat(cellB)) {
```
BudnikOleksii: It's a bad idea to check like this

BudnikOleksii (approved):
Good work 👍 
There is something wrong with the deployment, but by code it should be ok. But do better checks in the future

PR #1050, opened 2024-12-16
add solution

nester7an:


etojeDenys (changes requested):
pls, don't hesitate to ask for some help with demo link in the chat and then re-request the review
image

PR #1205, opened 2025-04-25
task 1

apr1ori:


src/scripts/main.js
```diff
@@ -1,3 +1,79 @@
...
+salary.addEventListener('click', () => {
+  const newRows = Array.from(rows);
+  const infos = [];
+
+  newRows.forEach((item) => {
+    const strVal = item.cells[3].textContent;
+    const numVal = Number(strVal.replace(/[^0-9]/g, ''));
+
+    infos.push({ numVal, strVal });
+  });
```
etojeDenys, line 70: it's better to rewrite it using the map method

src/scripts/main.js
```diff
@@ -1,3 +1,79 @@
...
+  }
+});
+
+age.addEventListener('click', () => {
+  const newRows = Array.from(rows);
+  const infos = [];
+
+  newRows.forEach((item) => {
+    infos.push(item.cells[2].textContent);
+  });
```
etojeDenys, line 50: same here

src/scripts/main.js
```diff
@@ -1,3 +1,79 @@
...
+
+  const sortedInfos = infos.sort((a, b) => {
+    return a.localeCompare(b);
+  });
+
+  for (let i = 0; i < newRows.length; i++) {
+    newRows[i].cells[1].textContent = sortedInfos[i];
+  }
+});
```
etojeDenys, line 42: try to combine all these listeners into one

etojeDenys (changes requested):

PR #1215, opened 2025-05-08
implement table sorting in ASC order, by clicking on table headers

galina-palyukh:


galina-palyukh:
@mateacademy-ai-mentor 
My task was to implement table sorting in ASC order, by clicking on table headers.
So I shouldn't need to add descending order

PR #1337, opened 2025-10-15
add task solution

Vall-Re:
- [DEMO LINK](https://Vall-Re.github.io/js_sort_table_DOM/)


Vall-Re:
- [DEMO LINK](https://Vall-Re.github.io/js_sort_table_DOM/)
 - 
What's fixed:

Initialization after DOM is ready
→ guarantee that thead and tbody exist, will not be null.

Correct click handling
→ e.target.closest('th') allows clicks on children of <th>.

Checking index and cell presence
→ if the structure changes (colspan, multi-line header), the code will not crash.

Stable number parsing
→ keep minus (-), remove commas, fallback for NaN.

Safe tbody clearing
→ use textContent = '' to simply clear the DOM (similar to innerHTML, but a little safer and faster).

Case-insensitive text sorting
→ localeCompare with { sensitivity: 'base' } (ignores case).

